Issue #102, opened 2026-06-19
Upgraded installs do not activate newly added auditor/default persona updates


HaoxuanLiTHUAI:
## Summary

BrainPilot has a built-in `auditor` persona and updated PI persona logic in source, but upgraded/local installs with an existing `bp_template/` may continue running old agent templates. In that state, the auditor exists in source but is not effectively activated in the product path.

## Why this matters

Users can upgrade to a version that includes the auditor agent, but their existing data directory may still contain:

- no `bp_template/agents/auditor/` template
- an older `bp_template/agents/principal/prompt.md` that never tells PI to use `auditor`
- an older PI manifest/template that does not expose the current default tool expectations such as `ask_user`

Because the runtime prefers `bp_template/agents/<name>/prompt.md` over the built-in persona, the old on-disk PI prompt shadows the improved source prompt. The result is that auditor/fabrication-check behavior is present in code but absent or inconsistent for upgraded installs.

## Repro outline

1. Start from a data dir scaffolded before `auditor` was added or before the PI audit/clarification prompt was updated.
2. Upgrade BrainPilot and run `brainpilot up` using the same data dir.
3. Inspect `bp_template/agents/`.
4. Ask PI for a task that produces numeric/file/reference claims through experts.

## Observed

- `brainpilot up` only scaffolds when `bp_template/` is absent.
- Existing `bp_template/agents/principal/prompt.md` remains unchanged and shadows the built-in updated PI persona.
- Existing data dirs may not get `bp_template/agents/auditor/` at all.
- PI may not call `auditor` for expert outputs or final deliverables, even though source contains an auditor persona.

## Expected

Upgraded installs should preserve user edits while still making new built-in agents and critical default persona updates available. Possible approaches:

- add a non-destructive template sync/migration that creates missing built-in agent directories such as `auditor`
- track template/persona versions and surface when local prompts are stale
- store built-in defaults separately from user overrides so users can diff/accept updates
- at minimum, warn in the UI/CLI when a local prompt shadows a newer built-in default

## Relevant code paths

- `packages/runtime/src/session-manager.ts`: runtime prefers on-disk `bp_template/agents/<name>/prompt.md`
- `packages/cli/src/commands/up.ts`: only calls `scaffold()` if `isScaffolded(dataDir)` is false
- `packages/cli/src/scaffold.ts`: `writeIfAbsent()` preserves existing files, which is good for user edits but leaves no migration path for newly added built-ins

[...]

PR #170, opened 2026-06-25
fix(cli,backend): launchable from npm install + thread --dir to runtime (#169)

### LINKED ISSUE #169: [Bug]: installed CLI won't launch (repo-cwd guard) + `--dir` splits the data dir (runtime ignores it)
## Summary

A fresh `npm install -g @brainpilot/app` cannot be launched the way users expect, and the documented workaround (`--dir`) writes data to two different places. Two distinct bugs, both reproduced on `@brainpilot/app@0.0.7` (installed) and still present in source at HEAD (`0.0.8`, testbed `71fe5ad`).

A user hit this as: *"installed the package, launching shows an internal error."* The first wall they hit is bug #1; bug #2 then bites anyone who follows the guard's own `--dir` advice.

---

## Bug 1 — `assertRepoCwd` makes a globally/locally installed CLI unlaunchable

`bin.js` runs `assertRepoCwd()` before commander, and the only non-escape-hatch pass condition is:

```ts
// packages/cli/src/repo-mode.ts:92-93
const expected = resolve(cwd, "packages/cli/dist/bin.js");
if (samePath(expected, binPath)) return;
```

This requires the **running `bin.js`** to be exactly `<cwd>/packages/cli/dist/bin.js`. That path only exists when you `git clone` the monorepo and run from its root. Under any npm install the bin lives at:

```
<prefix>/node_modules/@brainpilot/app/dist/bin.js
```

— there is no `packages/cli/` layer, because the published package is a single extracted workspace, not the repo skeleton (`package.json` `directory: "packages/cli"` is just metadata). So `samePath(expected, binPath)` is **always false** for installed users, and the guard exits before *any* command runs — including `--help` and `--version`.

### Repro
```console
$ npm install -g @brainpilot/app
$ cd ~ && brainpilot --help
✗ brainpilot 必须在仓库根目录运行 (当前 cwd 不是 BrainPilotPi 仓库根)
    cwd      = /Users/me
    bin.js   = /Users/me/.nvm/.../node_modules/@brainpilot/app/dist/bin.js
...
```
`brainpilot up`, `brainpilot status`, etc. all fail identically. The bin is genuinely on `PATH` and functional — the guard is the only thing stopping it.

### Why it's a contradiction
The package is published (`publishConfig.access: public`), advertises `bin: brainpilot / bnpt`, and has a `homepage`/README that read like an installable CLI — but the launch contract hard-codes "must run from a cloned repo root." Those can't both be true.

### Workarounds (confirmed working)
Any one of the escape hatches lets it run:
```bash
BP_ALLOW_FOREIGN_CWD=1 brainpilot up
brainpilot up --dir ~/brainpilot-data      # but see Bug 2
BP_DATA_DIR=~/brainpilot-data brainpilot up
```

### Suggested fix
Detect "running from inside `node_modules`" and allow it (global + local installs are legitimate non-repo launches); keep the `packages/cli` check only for the from-source dev case. Solve the data-dir-drift concern (the original motivation, #102) with a stable default like `~/.brainpilot` instead of tying it to cwd. If clone-and-run is truly the *only* supported mode, then don't ship an installable bin to npm (or ship a stub bin that prints "please clone the repo") and document "do not `npm install -g`."

---

## Bug 2 — `--dir` is not propagated to the runtime; data dir splits in two

`startServer` passes `dataDir` to the app but **not** to the orchestrator:

```ts
// packages/backend-core/src/server.ts:41-42
const orchestrator = options.orchestrator ??
  createOrchestrator({
    local: options.stdioInherit ? { stdioInherit: true } : undefined,
  });   // <-- options.dataDir is dropped here
```

So in the factory, `dataDir` falls back to the default:

```ts
// packages/backend-core/src/create-orchestrator.ts:52
const dataDir = options.local?.dataDir ?? env.BP_DATA_DIR ?? "./brainpilot";
```

The orchestrator then spawns the runtime child with `BP_DATA_DIR=./brainpilot` (relative to cwd), and the runtime materializes skills under `<cwd>/brainpilot/bp_template/skills` — **ignoring `--dir`**. Meanwhile the backend's own scaffold (providers/mcp/agents) correctly went to the `--dir` location. Result: one launch, two data dirs.

This only manifests once the runtime actually boots (it's lazy — first request that needs an agent), so a bare page load won't show it.

### Repro
```bash
mkdir /tmp/bp-flag && cd /tmp/bp-flag
BP_MOCK=1 brainpilot up --dir /tmp/bp-flag/data --port 19096 --no-open &
# create a session + send a message to force the runtime to spawn
sid=$(curl -s -XPOST localhost:19096/api/sessions -d '{}' | jq -r .id)
curl -s -XPOST localhost:19096/api/sessions/$sid/messages -d '{"content":"hi"}'

ls /tmp/bp-flag/data/bp_template/skills    # backend scaffold (correct dir)
ls /tmp/bp-flag/brainpilot/bp_template/skills  # runtime skills LEAKED here
```

Log lines make the divergence explicit:

| launch | runtime log |
|---|---|
| `--dir /tmp/bp-flag/data` | `copied → brainpilot/bp_template/skills` ❌ (relative, wrong) |
| `BP_DATA_DIR=/tmp/bp-env/data` | `copied → /tmp/bp-env/data/bp_template/skills` ✅ |

Using the **`BP_DATA_DIR` env var works** (it flows through `create-orchestrator.ts:52`'s `env.BP_DATA_DIR` branch); only the **`--dir` flag** is dropped. That asymmetry is the tell.

### Impact
- Skills, MCP config, and session/workspace persistence land in different roots, so the runtime can't see what the CLI scaffolded.
- A stray `./brainpilot/` dir is created next to wherever the user `cd`'d — the exact drift the `repo-mode` guard was trying to prevent, reintroduced through the back door.

### Suggested fix
Thread `dataDir` through to the orchestrator:
```ts
createOrchestrator({
  local: {
    dataDir: options.dataDir,
    ...(options.stdioInherit ? { stdioInherit: true } : {}),
  },
});
```
(or have `createOrchestrator` accept top-level `dataDir`). With that, `LocalProcessOrchestrator.buildEnv` sets `BP_DATA_DIR` to the resolved `--dir` and the runtime child materializes into the same root.

---

## Environment
- `@brainpilot/app@0.0.7` (npm global), Node v24.13.0, macOS (darwin 25.4.0)
- Source verified at testbed `71fe5ad` (v0.0.8) — both bugs still present.

## Relationship between the two
They compound: bug #1 pushes users toward `--dir`/`BP_DATA_DIR` as the way to launch outside a clone, and bug #1's own error message recommends `--dir` — but `--dir` then triggers bug #2. The env-var workaround avoids both.

HaoxuanLiTHUAI:
Fixes #169.

Two install-blocking bugs found while reproducing #169, each with a green regression test.

## Bug 1 — installed CLI was unlaunchable (`repo-mode` guard)

`assertRepoCwd` only passed when the running `bin.js` equalled `<cwd>/packages/cli/dist/bin.js` — a path shape that exists only in a cloned repo. A global/local `npm install` ships the CLI at `<prefix>/node_modules/@brainpilot/app/dist/bin.js`, so the guard exited (code 2) before *any* command (even `--help`/`--version`) could run.

**Fix:** skip the guard when `bin.js` lives inside a `node_modules/` tree (a legitimate non-repo launch). The repo-root contract is kept intact for genuine from-source runs. The #102 "configs update with `git pull`" rationale doesn't apply to installed users — they have no repo to pull.

## Bug 2 — `--dir` dropped on the way to the orchestrator (data dir split)

`startServer` built the orchestrator **without** `options.dataDir`, so the spawned runtime fell back to `BP_DATA_DIR=./brainpilot` (cwd-relative) and materialized skills / persisted sessions there, while the backend scaffold wrote to the requested `--dir`. One launch ended up with two data roots. (The `BP_DATA_DIR` *env var* already worked — only the `--dir` *flag* was lost, via `create-orchestrator.ts`'s `env.BP_DATA_DIR` branch.)

**Fix:** extract `buildServerOrchestrator(options)` (pure, exported) and pass `local.dataDir` through to the local orchestrator.

## Why the existing tests didn't catch these

- `packages/cli/src/repo-mode.test.ts` only modelled the *from-source* `packages/cli/dist/bin.js` layout — it never built a `node_modules` install path, so it froze the buggy behavior as "correct."
- `packages/backend-core/test/create-orchestrator.test.ts` always fed `dataDir`/`BP_DATA_DIR` straight into the factory, never exercising the `startServer` → `createOrchestrator` caller seam where the drop happened. backend-core had no `server.test.ts`, and `up.test.ts` mocks `startServer` entirely.

## Tests added (both proven RED before the fix, GREEN after)

- `repo-mode.test.ts`: global + local `node_modules` layouts launch from any cwd; a from-source run from the wrong cwd still rejects (contract preserved).
- `backend-core/test/server.test.ts` (new): `buildServerOrchestrator` threads `dataDir` into the local runtime env in both stdio modes; injected-orchestrator short-circuit; `BP_DATA_DIR` env fallback.

## Verification

- `repo-mode.test.ts` and `server.test.ts` confirmed RED with the product fix reverted, GREEN with it applied.
- Full `packages/backend-core` + `packages/cli` suites: **282 passing**. `tsc -b` typecheck clean.
- End-to-end against the built `dist`:
  - `up --dir /tmp/x/data` → skills land in `/tmp/x/data/bp_template/skills`, **nothing** leaks to `./brainpilot` (previously split).
  - install-layout `bin.js` launches from a foreign cwd; from-source wrong-cwd still shows the repo-root guard.

## Out of scope (flagged for maintainers)

This PR only unblocks launch + fixes the split. The deeper design question from #169 — whether the default data dir should move to a stable `~/.brainpilot` rather than `<cwd>/brainpilot` — is left to the maintainers, since it changes user-visible behavior beyond the bug.

### DIFF
--- a/packages/backend-core/src/index.ts
+++ b/packages/backend-core/src/index.ts
@@ -8,7 +8,7 @@ export const BACKEND_NAME = "@brainpilot/backend-core";
 export { createApp } from "./app.js";
 export type { CreateAppOptions } from "./app.js";
 
-export { startServer } from "./server.js";
+export { startServer, buildServerOrchestrator } from "./server.js";
 export type { StartServerOptions, RunningServer } from "./server.js";
 
 export { createOrchestrator } from "./create-orchestrator.js";
--- a/packages/backend-core/src/server.ts
+++ b/packages/backend-core/src/server.ts
@@ -29,6 +29,30 @@ export interface RunningServer {
   stop: () => Promise<void>;
 }
 
+/**
+ * Build the orchestrator for a server from its options. Exposed (and pure) so
+ * the dataDir wiring is unit-testable without binding a socket.
+ *
+ * Issue #169: the local orchestrator MUST receive `options.dataDir` — it spawns
+ * the runtime child with `BP_DATA_DIR=<dataDir>`, and the runtime materializes
+ * skills / persists sessions under that root. Dropping it here made the runtime
+ * fall back to `./brainpilot` (relative to cwd) while the backend scaffold wrote
+ * to the requested `--dir`, splitting one launch across two data dirs.
+ */
+export function buildServerOrchestrator(
+  options: StartServerOptions = {},
+): Orchestrator {
+  return (
+    options.orchestrator ??
+    createOrchestrator({
+      local: {
+        dataDir: options.dataDir,
+        ...(options.stdioInherit ? { stdioInherit: true } : {}),
+      },
+    })
+  );
+}
+
 export async function startServer(
   options: StartServerOptions = {},
 ): Promise<RunningServer> {
@@ -36,11 +60,7 @@ export async function startServer(
   // 127.0.0.1 is the safe default for local/CLI use; containers set BP_HOST=0.0.0.0
   // so the published port (DNAT'd to the container IP) can reach the server.
   const hostname = options.hostname ?? process.env.BP_HOST ?? "127.0.0.1";
-  const orchestrator =
-    options.orchestrator ??
-    createOrchestrator({
-      local: options.stdioInherit ? { stdioInherit: true } : undefined,
-    });
+  const orchestrator = buildServerOrchestrator(options);
 
   const app = createApp({
     orchestrator,
--- a/packages/backend-core/test/server.test.ts
+++ b/packages/backend-core/test/server.test.ts
@@ -1,0 +1,53 @@
+import { describe, expect, it } from "vitest";
+import { buildServerOrchestrator } from "../src/server.js";
+import { LocalProcessOrchestrator } from "../src/local-orchestrator.js";
+import type { Orchestrator } from "../src/orchestrator.js";
+
+/**
+ * Regression coverage for issue #169 — `startServer` dropped `dataDir` when
+ * constructing the orchestrator, so the runtime child it spawns fell back to
+ * `./brainpilot` (relative to cwd) instead of the requested data dir. The
+ * orchestrator construction is now a pure, exported function so this seam is
+ * testable without binding a socket.
+ *
+ * The `create-orchestrator.test.ts` suite already covers the factory in
+ * isolation; what was missing was the *caller* wiring — that `options.dataDir`
+ * actually reaches the local orchestrator's `BP_DATA_DIR`.
+ */
+describe("buildServerOrchestrator", () => {
+  it("threads dataDir through to the local runtime env (#169)", () => {
+    const orch = buildServerOrchestrator({
+      dataDir: "/data/xyz",
+      // foreground/CLI mode — exercises the stdioInherit branch too.
+      stdioInherit: true,
+    });
+    expect(orch).toBeInstanceOf(LocalProcessOrchestrator);
+    const env = (orch as LocalProcessOrchestrator).buildEnv();
+    expect(env.BP_DATA_DIR).toBe("/data/xyz");
+  });
+
+  it("threads dataDir through in detached (non-stdioInherit) mode too", () => {
+    const orch = buildServerOrchestrator({ dataDir: "/data/detached" });
+    expect(orch).toBeInstanceOf(LocalProcessOrchestrator);
+    const env = (orch as LocalProcessOrchestrator).buildEnv();
+    expect(env.BP_DATA_DIR).toBe("/data/detached");
+  });
+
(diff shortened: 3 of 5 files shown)


HaoxuanLiTHUAI:
Superseded by #172 — retargeted from the now-frozen `development` branch to `main` (sole trunk since ~6/23). Same fix + tests.

Issue #157, opened 2026-06-23
[Bug]: Windows node_modules path (with username) leaks in agent error — redact() regex is forward-slash only


HaoxuanLiTHUAI:
### What happened

On Windows, agent error messages that carry an absolute `node_modules` doc path are **not redacted**, so the user's local absolute path — **including their Windows username** — leaks into the chat error bubble and is persisted to `.bp/<sid>/events.jsonl`.

`normalizeAgentError()` → `redact()` strips local-path leakage with two regexes that hardcode a **forward-slash** `node_modules/`:

```ts
// packages/runtime/src/agent-error.ts
if (/^\/?\S*node_modules\/\S+$/.test(t)) return false;   // line 157 — drop a path-only line
line.replace(/\/\S*node_modules\/\S+/g, "").trimEnd(),   // line 162 — scrub inline path
```

On Windows the Pi SDK guidance error contains a native **backslash** absolute path
(`C:\Users\<user>\...\node_modules\@earendil-works\pi-coding-agent\docs\providers.md`),
which neither regex matches → the full path is shown and stored verbatim.

### Reproduce (deterministic, no Windows machine needed — it's a pure function)

```bash
npm run build --workspace @brainpilot/runtime
node -e 'import("./packages/runtime/dist/agent-error.js").then(m=>{
  const win   = "No model selected.\n  C:\\Users\\alice\\AppData\\Roaming\\npm\\node_modules\\@earendil-works\\pi-coding-agent\\docs\\providers.md";
  const posix = "No model selected.\n  /home/u/app/node_modules/@earendil-works/pi-coding-agent/docs/providers.md";
  console.log("POSIX  :", JSON.stringify(m.normalizeAgentError(posix).message));
  console.log("WINDOWS:", JSON.stringify(m.normalizeAgentError(win).message));
})'
```

Observed:
- **POSIX** → `"No model selected."` (path scrubbed ✅)
- **WINDOWS** → `"No model selected.\n  C:\\Users\\alice\\...\\node_modules\\...\\providers.md"` (full path **+ username `alice`** leaked ❌)

### Expected

Regardless of path separator, absolute `node_modules` paths should be redacted; in particular a Windows host absolute path (with username) must never reach the displayed error or `events.jsonl`. The semantic message head (`No model selected.`) should be kept.

### When it triggers

[...]

PR #173, opened 2026-06-25
fix(cli): lock `brainpilot up` to local mode unless mode is explicit

GTC2333:
## Problem

Users launching from source (`npm run bp -- up`) could **silently land in static (sandbox) mode**. The orchestrator mode was resolved purely from the environment, so a stray `BP_RUNTIME_URL` left over from a `docker compose` session flipped a source launch into static mode with no indication. The only symptom was a confusing `static runtime did not become healthy ... is the sandbox container up?` timeout — which points users at fixing a sandbox they never meant to use.

Root cause: `resolveOrchestratorMode()` precedence is `BP_ORCHESTRATOR` → `BP_RUNTIME_URL` (→static) → `BP_MODE` → local, and `brainpilot up` neither overrode it nor surfaced the result.

## Fix (1 + 2 + 3)

1. **Visibility** — `up` prints a mode banner on every launch: `mode=local` (dim) or, for static/docker, a yellow warning that names the trigger (`--mode` vs `BP_ORCHESTRATOR`) and echoes `BP_RUNTIME_URL`. The switch is never silent again.
2. **Lock to local unless explicit** — new `resolveUpMode()` defaults to `local` and honors only `--mode` and an explicit `BP_ORCHESTRATOR`; it **deliberately ignores `BP_RUNTIME_URL` / `BP_MODE`** so leftover env can't hijack a source launch. The resolved mode is passed explicitly via `StartServerOptions.mode`; the detached path pins `BP_ORCHESTRATOR` so the background server agrees. Adds `--mode <local|static|docker>`.
3. **Better error** — the static-orchestrator timeout now splits the two cases (real Docker deploy vs. stray env on a source launch) and gives the exact escape hatch (`brainpilot up --mode local` / `unset BP_RUNTIME_URL BP_ORCHESTRATOR`).

## Safety

Backend `startServer` env resolution is **unchanged**, so Docker compose — which selects static via `BP_RUNTIME_URL` on `node server.js` — is unaffected.

## Tests

+11 tests (`resolveUpMode`, `up` banner/mode passthrough, `--mode` flag + validation). All **572 non-web tests pass**; typecheck clean. Verified end-to-end:
- `BP_RUNTIME_URL=… up` (no `--mode`) → stays `mode=local`, starts normally ✅
- `BP_RUNTIME_URL=… up --mode static` → warns + names trigger ✅

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README-zh.md
+++ b/README-zh.md
@@ -75,17 +75,29 @@ npm install -g @brainpilot/app
 
 这会安装 `brainpilot` 命令（`bnpt` 是同一命令的内置短别名）。
 
-### 2. 初始化
+### 2. 配置模型服务商
+
+BrainPilot **不绑定 Anthropic** —— 支持多种服务商协议：**Anthropic Messages**、**OpenAI
+Completions**、**OpenAI Responses** 和 **Azure OpenAI Responses**。按你的模型/网关所用协议任选。
+
+推荐做法是启动后用 **Web Settings UI**：打开 **Settings → Providers（服务商）**，添加一个服务商
+（base URL / key / 协议 / 模型列表），它会自动帮你写入配置。缺少 Key 不再阻塞启动：
+**`brainpilot up` 照常启动**，所以你完全可以先跳过、在浏览器里配置服务商。
+
+<details>
+<summary><b>更想用命令行？在 init 时生成配置</b></summary>
 
 ```bash
+# Anthropic（默认协议）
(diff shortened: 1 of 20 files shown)


GTC2333:
Superseded by a clean branch based directly on `main` (the original branch was cut from a local `development` that carried an unmerged #157 and diverged from `main`, producing conflicts). New PR: see below.

PR #367, opened 2026-07-24
[P0] Retry transient provider failures and persist model health

### LINKED ISSUE #364: [Bug]: Provider Test shows unknown for a working OpenAI Chat Completions provider
## Bug Description

In **Settings → Providers**, clicking **Test** can leave the status as `unknown` for a working OpenAI Chat Completions provider. The same provider and model work in real BrainPilot sessions.

Tested configuration:

- Protocol: `OpenAI Completions` (`openai-completions`)
- Base URL: `https://llmapi.paratera.com/v1`
- Model: `Qwen3.6-27B`
- API key: valid and intentionally omitted from this issue

## Steps to Reproduce

1. Open **Settings → Providers**.
2. Add a provider using the configuration above and a valid API key.
3. Save and select the provider.
4. Confirm that a normal BrainPilot session can call the model successfully.
5. Return to the Provider page and click **Test**.

## Expected Behavior

The test should report and persist `healthy` for the provider and the tested model.

If the button only tests provider-level connectivity, the UI should clearly distinguish **provider healthy / model not tested** from a generic `unknown` result.

## Actual Behavior

The Provider page displays `unknown` even though the provider works in a real session.

## Deployment Method

Other / exact deployment version not yet captured.

## BrainPilot Version

Unknown. The behavior was reported on 2026-07-24.

## Independent Verification

The gateway was checked directly with the same valid credential, without exposing it:

- `GET https://llmapi.paratera.com/v1/models` returned HTTP 200 in about 70 ms and included `Qwen3.6-27B`.
- `POST https://llmapi.paratera.com/v1/chat/completions` with the same minimal request used by the current protocol-aware probe returned HTTP 200 in about 2.6 s, with `object: chat.completion` and an `OK` response.

This rules out the base URL, API key, model ID, and Chat Completions compatibility as the immediate cause.

## Likely Code Path

On current `main`, `POST /provider/profiles/:id/test` persists the top-level probe result, but explicitly leaves `model_health` empty because per-model probing is marked as future work. The Web UI maps a missing model status to `unknown`.

This makes the Test result ambiguous: the provider probe may succeed while the visible model status remains unknown. If the reported `unknown` is the top-level provider status, the test response or persisted health value is also not reaching the UI correctly.

Related issues #55 and #69 addressed real connectivity testing and persistence of top-level provider health, so this appears to be a remaining model-health/UI gap or a regression around those paths.

### LINKED ISSUE #365: [P0] Principal stops on transient provider errors instead of retrying
## Summary

A transient provider error currently terminates the principal run immediately. There is no host-level retry for the direct principal prompt path, so the user sees an error and the agent stops even when the same request succeeds moments later.

This should be treated as P0 because a short-lived upstream failure causes an unrecoverable task interruption.

## Production evidence

On 2026-07-24 at 17:11:45 UTC+8, a principal run failed with:

```
HTTP 400
{
  "message": "The request is invalid: {\"message\":\"invalid request error trace_id: 957c58f4b25567153aa3985bad9e3f72\",\"type\":\"invalid_request_error\"}\n. Please check the request body, required fields, and request format.",
  "type": "invalid_request_error",
  "param": "",
  "code": null
}
```

The same simple user request succeeded roughly two minutes later without any configuration change.

Production logs for the same day contain about 31 errors with this identical generic 400 shape across 6 users and 8 sessions. They were routed through the OpenAI Completions-compatible provider to `deepseek-v4-pro`. The evidence strongly suggests that at least some of these generic upstream 400 responses are transient rather than deterministic request validation failures.

## Current behavior

- The direct principal path calls `agent.prompt(content)` and clears the run when it returns; there is no session-manager retry.
- `MasAgent.prompt` converts the provider failure into agent error state, so an outer promise `.catch()` is not sufficient to recover.
- Expert delivery has a separate bounded retry path, but the principal path does not.
- The underlying SDK retries typical transient classes such as 429, 5xx, timeouts, and network failures, but not this generic 400 `invalid_request_error`.

Relevant code:

- `packages/runtime/src/session-manager.ts` — direct principal prompt and expert delivery retry paths
- `packages/runtime/src/mas-agent.ts` — prompt error isolation/state transition
- `packages/runtime/src/provider-config.ts` — provider/model resolution

## Expected behavior

Introduce one unified, bounded retry policy for principal and expert model calls.

Required policy:

1. Retry up to 5 times after the initial attempt (6 total attempts).
2. Increase the wait interval after each failure using bounded exponential backoff (2s, 4s, 8s, 16s, 32s).
3. Retry known transient errors: 429, retryable 5xx, timeout/network errors, and this upstream generic 400 shape when it contains `invalid_request_error`, a trace ID, and no actionable `param`/`code`.
4. Do not retry authentication/authorization failures (401/403) or deterministic validation errors that identify a concrete invalid field.
5. Retry a whole turn only before meaningful assistant output or a side-effectful tool call has occurred, to prevent duplicate output and duplicated side effects.
6. Abort/interrupt must immediately cancel backoff and any pending retry.
7. While retrying, reuse the existing `principal is working` bubble and show the current retry state (for example, `Retrying 2/5 in 4s`). Do not add a separate error bubble for each attempt.
8. If retries are exhausted, preserve the upstream trace ID and retry count in the final error and telemetry.

## Acceptance criteria

- [ ] A transient principal provider failure is retried automatically up to 5 times and can recover without user action.
- [ ] Wait intervals increase between retries and remain bounded.
- [ ] The `principal is working` bubble visibly reports retry progress/countdown.
- [ ] Principal and expert calls use the same retry classification and backoff policy.
- [ ] A retry cannot duplicate assistant messages or side-effectful tool execution.
- [ ] Interrupting a run cancels pending retries.
- [ ] Deterministic 4xx and auth errors fail immediately.
- [ ] Retry attempts, classification, provider/model, and upstream trace ID are observable in logs/telemetry.
- [ ] Unit/integration coverage includes: transient generic 400 then success, all 5 retries exhausted, increasing backoff, 429/5xx, deterministic 400, UI retry state, and interrupt during backoff.

## Additional production finding

The affected session referenced a stale `preset-custom/custom-model-a1` profile. Because that profile no longer existed, runtime resolution fell back to the currently selected profile/model. This is not sufficient to explain the transient behavior above, but provider-resolution fallback should be included in telemetry so future incidents show the effective provider/model clearly.

Related: #167, #97

HaoxuanLiTHUAI:
## Summary

- retry provider failures up to 5 times inside the same Pi LLM turn with increasing 2s/4s/8s/16s/32s backoff
- recognize the narrow transient HTTP 400 shape observed in production (`invalid_request_error` + trace ID + no actionable param/code) while leaving deterministic 4xx/auth failures non-retryable
- keep transient attempts out of chat error bubbles and show retry progress in the existing `principal is working` toast
- persist the protocol-aware provider probe result for the model that was actually tested; return explicit `unknown` health for untested configured models

The same-turn Pi retry path removes the failed assistant message and continues the provider call before a subsequent tool can run, avoiding duplicate user messages and side-effectful tool calls. Interrupt remains abortable through Pi's existing retry abort controller.

## Tests

- `npm test` — 71 files / 789 tests passed
- `npm test -w @brainpilot/web` — 48 files / 391 tests passed
- `npm run typecheck`
- `npm run build --workspace @brainpilot/protocol --workspace @brainpilot/runtime --workspace @brainpilot/backend-core --workspace @brainpilot/web`

Closes #365
Closes #364


GTC2333 (changes requested):
Found one blocking lifecycle bug:

When the user interrupts during Pi’s retry backoff, Pi emits `auto_retry_end { success: false, finalError: "Retry cancelled" }`. The new handler stores that as `pendingProviderError`; after `session.prompt()` unwinds, `runPrompt()` calls `recordError()` and leaves the agent in `error`. A normal user Stop therefore creates a red provider-error bubble and an error run/state, despite being an intentional cancellation.

Please distinguish retry cancellation/abort from retry exhaustion, and add an interrupt-during-backoff regression test asserting no error system message and a non-error terminal agent state. The retry classifier, same-turn retry behavior, health persistence, protocol mapping, and UI progress path otherwise look coherent.

HaoxuanLiTHUAI:
已修复本轮 review 指出的 Stop / Retry 生命周期问题（commit d53b237）：

- 显式区分用户主动 Stop 与 provider retry 耗尽
- Stop 时清除 pending provider error，并忽略 `Retry cancelled`，最终回到 idle、run outcome 记为 aborted
- principal 与 delegated expert 两条路径均增加回归测试；delegated expert 不再继续 delivery retry / escalation
- 验证：完整测试 792 passed，typecheck、runtime build、git diff --check 均通过

@GTC2333 麻烦再帮忙 review 一下。

GTC2333 (approved):
Re-reviewed commit d53b237 and the updated branch. The interrupt-during-retry lifecycle bug is fixed for both principal and delegated expert paths, regression coverage is present, and fresh required CI passes on current main.

Issue #194, opened 2026-06-29
Refresh drops chat history: SSE-tail race truncates rehydrate (B1) + evicted-session history 404 (B2)


GTC2333:
## Summary

After a browser refresh, the chat transcript is lost or truncated even though
the backend still has the full event log on disk (`events.jsonl`). Two distinct
severities are observed:

- **B1 — only the last few messages survive** a refresh (most sessions).
- **B2 — one session lost its entire transcript** after refresh.

In all cases the data is intact server-side ("后端都可以看到"), so the loss is in
the **read/replay path**, not persistence.

## B1: refresh keeps only the most recent messages (rehydrate skipped by SSE race)

Root cause is a **seed/SSE race** in `SessionContext`:

1. On refresh, `connectSession` attaches SSE immediately; the runtime replays its
   **recent ring buffer** (`EventBus` `maxBuffer`, a small tail) on connect.
2. The SSE drain effect reduces those few recent events into
   `messagesBySession[sid]`, so its length becomes `> 0`
   (`SessionContext.tsx:840`).
3. The async history rehydrate (`limit=0`, full log) lands later, but its seed is
   guarded by `if ((current[sessionId]?.length ?? 0) > 0) return current`
   (`SessionContext.tsx:299`) — so the **full history is dropped** and only the
   SSE tail remains.

The `limit=0` → full-log request itself is correct end-to-end (web sends
`?limit=0`; backend forwards the query verbatim; runtime treats `limit<=0` as the
whole log). The bug is purely the client-side seed-vs-SSE ordering: history must
**merge** with whatever SSE already deposited (the reducers already dedupe by
messageId/toolCallId), not bail when the list is non-empty.

## B2: a whole session comes back empty (history 404 after eviction)

Same root cause as #165: `SessionManager.readEventHistory` starts with
`if (!this.sessions.has(sessionId)) return undefined`, so once a session is
evicted from memory (idle reaper / runtime restart) the history endpoint returns
404 even though `events.jsonl` is still on disk (`evictSession` flushes and
deletes the in-memory entry but never removes the file). The web client's
`getHistory` maps any non-200 to an empty envelope (`utils/api.ts`), so the whole
transcript silently disappears. `bpDir(sid)` depends only on the session id, so a
disk-backed read is possible without an in-memory entry.

## Fix direction

- **B1**: make the rehydrate seed **merge into** the existing list through the same
  reducers instead of skipping when `length > 0`. Re-fold the full history over
[...]

PR #384, opened 2026-07-27
fix: add authoritative tool lifecycle and scoped script interruption

### LINKED ISSUE #380: [Bug][UX] Running Bash timer resets on re-entry and Stop leaves stale script UI
## Bug description

There are two related state/lifecycle problems with the Running Scripts UI:

1. Every time the user enters or reloads the session UI, the elapsed time shown for an already-running Bash script starts again from zero instead of continuing from the script's actual start time.
2. Clicking **Stop** does not settle the interrupted run correctly:
   - the agent does not provide or reach a clear terminal response/state;
   - the Bash script remains visible in the **Running Scripts** panel;
   - clicking **Stop** repeatedly only adds repeated interruption notices;
   - the script UI and Stop control remain present even though interruption was requested.

This appears to be a regression or an incomplete case related to #327.

## Steps to reproduce

1. Start a task that launches a long-running Bash command.
2. Leave the session page while the command is still running.
3. Re-enter or reload the session.
4. Observe the Bash elapsed-time counter.
5. Click **Stop** in the Running Scripts panel.
6. Wait for the interrupt request to settle.
7. Click **Stop** again one or more times.

## Expected behavior

- The elapsed-time counter is derived from the persisted script start timestamp and continues across navigation, reload, and session rehydration.
- A successful interrupt terminates the running script and settles the agent/run state.
- The Running Scripts card and Stop control disappear once the authoritative terminal event is received.
- The UI emits at most one deterministic interruption acknowledgement.
- Repeated Stop clicks are disabled or idempotent while interruption is pending.
- The composer becomes usable again after the interruption settles.

## Actual behavior

- Re-entering the UI resets the displayed Bash duration to zero.
- Stop requests do not clear the running script from the frontend.
- The agent does not visibly settle/respond after interruption.
- Repeated clicks produce duplicate interruption notices while the stale script card remains visible.

## Impact

The UI reports a script as running after the user has attempted to stop it, so users cannot tell whether work is still executing. The reset timer also makes long-running jobs appear newly started and hides their true duration.

## Suggested investigation

- Persist and hydrate each script's authoritative `startedAt` rather than initializing the timer when the React component mounts.
- Trace the interrupt lifecycle across runtime script termination, agent/run terminal state, SSE events, and the Running Scripts reducer.
- Ensure Stop is idempotent and disabled while an interrupt request is in flight.
- Add an integration test covering navigation/reload during a running script followed by interruption.

## Environment

- Deployment method: unknown
- BrainPilot version: unknown
- Node.js version: unknown
- Operating system: unknown

## Screenshot

The observed UI shows three duplicate interruption notices and a Bash card still reporting “1 script running” with the Stop control visible after repeated interruption attempts.

GTC2333:
Closes #380

## Summary

- make tool start/end timestamps and terminal status authoritative and persistent
- expose rich `activeTools` state while retaining the legacy active-tool id list
- override Pi’s built-in local Bash with its public factory and add per-tool cancellation
- add a scoped tool interrupt endpoint across runtime, backend proxy, and web client
- separate **Stop script** from **Stop task**, including idempotent pending states
- reconcile history and SSE monotonically so completed tools do not revive after refresh

## Behavior

- **Stop script** cancels only the selected local Bash call; Pi receives an error tool result and the agent continues the same turn
- **Stop task** aborts the whole task, clears pending work, does not start another provider run, and emits one deterministic acknowledgement
- MCP Bash remains visible but does not expose precise cancellation
- detached processes whose tool call has already returned remain out of scope

## Compatibility

All new event and state fields are optional. Legacy `TOOL_CALL_END` events remain valid, and old histories derive duration from start/end timestamps when possible.

## Verification

- full non-web workspace suite: 815 tests passed
- full web suite: 403 tests passed
- production builds passed
- `git diff --check` passed

## Review focus

- Pi Bash definition override and preservation of timeout/partial-output behavior
- natural completion versus cancellation races and single terminal emission
- history/SSE terminal monotonicity
- whole-task interrupt idempotency and acknowledgement behavior


Fxde42:
## Review

Overall direction and structure look good, and this does genuinely fix both symptoms in #380 (elapsed time derived from the persisted `TOOL_CALL_START._ts`, plus monotonic terminal reconciliation). I verified the key assumptions hold:

- `createBashToolDefinition`, `SettingsManager.getShellCommandPrefix()`, and `getShellPath()` all exist in pi-coding-agent 0.80.3, and the override mirrors how Pi builds its own builtin bash (`agent-session.js:1933-1943`).
- `customTools` really do override builtins by name (`agent-session.js`, `toolRegistry.set(tool.name, tool)` after builtins), so the Bash override takes effect.
- `camelizeKey` preserves `_ts`, so the `createdAt` timer fix survives normalization.
- Deterministic `tool-result:<id>` message ids are an improvement for rehydrate dedupe, and each agent gets its own controller map.

Three things I'd resolve before merge, then some follow-ups.

### Should fix before merge

**1. Cancelling a script discards Pi's partial output** — `packages/runtime/src/agent-factory.ts:402-406`

Pi's bash throws `Error(<output produced so far> + "Command aborted")` on abort (`dist/core/tools/bash.js:294-295`). The wrapper replaces that entire message with `"Command interrupted by user"`, so **the model loses everything the script printed**. That contradicts this PR's own stated goal of "preserving timeout/partial-output behavior", and `cancellable-bash.test.ts` currently locks the behavior in. Suggest appending the marker instead of replacing the message.

**2. `mergeRehydratedMessages` is no longer scoped to tool messages** — `packages/web/src/contexts/SessionContext.tsx:191-222`

For any id present in both live and history where the live copy is finalized (`streaming === false`), the merge does `{...saved, ...live}` — so the live copy wins on **every** field, including `content`. That inverts the "persisted history is the base" rule documented for #194-B1 in the same docblock. The existing "canonical" test only still passes because its fixture leaves `streaming` as `undefined`.

The real path this breaks: the SSE ring-buffer tail can seed a truncated assistant message (earlier deltas already evicted from the 500-event buffer), which is then finalized by `TEXT_MESSAGE_END` — the truncated content now overwrites the complete history. Suggest gating the terminal-merge branch on `saved.kind === "tool"`.

**3. Stop task now surfaces a false error banner** — `session-manager.ts:1789-1795` + `SessionContext.tsx:773-777`

The new no-activity guard makes `interrupted: false` a common return (the old code returned true whenever any agent existed). The web treats `!interrupted` as failure and shows a red "interrupt failed". So clicking Stop just as a run settles now produces an error. The runtime already returns `reason: "already_idle"` for exactly this case, but the client ignores it.

### Follow-ups

- **Timeout leaves the tool permanently `stopping`** — `mas-agent.ts:637-646`: the timeout branch never resets `tool.status` back to `running`. A second Stop then reports `already_finished` (because `RealAgentSession.interruptTool` returns false once the controller is aborted) while the script is in fact still running. The runtime also replies 504, so the web throws and shows a raw HTTP error rather than anything actionable.
- **Raw enum shown to the user** — `SessionContext.tsx:1468`: `setError(result.reason ?? …)` renders `not_cancellable` / `already_finished` verbatim; everything else in this file goes through `tg()`.
- **The `agentsReducer` stable-reference optimization is now dead** — `agentsReducer.ts:68`: the runtime always emits `active_tools` (even `[]`), so `activeTools === undefined` is never true. Every `agent_status_update` allocates a new agents array and triggers a re-render, which the file's own docblock explicitly sets out to avoid ("a no-op status change returns the same reference"). A content comparison would restore it.
- **Normal completion can report a tool failure** — `mas-agent.ts:449`: `finishDanglingTools()` runs after a normal `prompt()` resolution and settles leftovers with `isError: true`, which emits the `❌ <tool> 执行失败` system_message. Any Pi path that starts a tool without a matching end (blocked tool call, `kind: "immediate"` result) will now show the user a failure banner. Consider a quieter status for this reconciliation.
- **Panel authority fails closed** — `runningScripts.ts`: `selectActiveScripts` drops any streaming bash absent from `activeTools`. After a reload the authority comes from the 500-event ring replay / the last persisted `session_state`; if the tool-start `agent_status_update` has been evicted, a genuinely running script disappears along with its Stop button. Relatedly, the comment at `SessionContext.tsx:1030` claiming `session_state` "arrives via SSE first frame on reconnect" isn't accurate — `server.ts:263` only replays the ring buffer and never emits a fresh snapshot.

### Nits

- `AbortSignal.any([runSignal, …])` and `runSignal.aborted` in the catch both assume a non-null signal, but the SDK types it `AbortSignal | undefined` (`ToolDefinition.execute`) — the hand-written `PiSdk` interface hides that divergence. If Pi ever passes `undefined`, the throw happens inside the catch and masks the original error.
- Elapsed time moved from monotonic `performance.now()` to wall clock (`Date.now()` vs a server ISO `startedAt`), so it's now sensitive to client/server clock skew. Per-item elapsed is clamped with `Math.max(0, …)`, but the header's `formatElapsed(now - oldest)` is not.
- Shell settings are read once at factory time, whereas Pi rebuilds its builtin bash on settings reload — and since `customTools` always beat builtins, a mid-session settings change would leave a stale prefix in place.
- `eventTerminals.sort((a, b) => Date.parse(b.completedAt ?? "") - …)` is a NaN comparator when `completedAt` is missing; it happens to work for two elements but isn't guaranteed by contract.
- `activeToolExecutions` is now threaded through the entire web layer but never read — fine if it's intentionally there for external clients, just noting it.

GTC2333:
Addressed the review feedback in b3bc0e6.

### Fixed before merge

1. **Partial Bash output is preserved** — local cancellation now appends `Command interrupted by user` to Pi’s original error text instead of replacing it. The wrapper also accepts Pi’s optional run signal.
2. **History merge remains canonical for normal messages** — terminal precedence is now gated to tool messages, with a regression test for a finalized but truncated SSE assistant tail.
3. **`already_idle` is benign** — Stop task no longer shows a false failure when the run settles concurrently.

### Follow-ups addressed here

- Added localized, actionable messages for `not_cancellable` and `timeout`; HTTP 504 tool lifecycle bodies are parsed rather than surfaced as raw HTTP errors.
- Restored stable-reference behavior for repeated identical `activeTools` updates.
- Missing Pi tool-end reconciliation remains terminal but no longer emits a user-facing failure warning.
- SSE reconnect now ends ring replay with a fresh authoritative `session_state`, preventing a live script from disappearing when its earlier state frame was evicted.
- Clamped the panel header duration and made terminal timestamp sorting safe for missing dates.

I intentionally keep a timed-out cancellation in `stopping`: the abort signal has already been sent, but process termination is unknown, so changing it back to `running` would also be inaccurate. The UI now directs the user to **Stop task**.

Verification after the changes:

- non-Web workspace: **816 tests passed**
- Web: **406 tests passed**
- runtime and Web production builds passed
- `git diff --check` passed

GTC2333:
Rebased onto the latest `main` after the review fixes. The review-fix commit is now `58bbe78` (same changes as the previously referenced `b3bc0e6`).

Fxde42 (approved):

PR #387, opened 2026-07-27
feat(mcp): preset BYOK + read-only presets (#377)

### LINKED ISSUE #377: [Feature] Hosted MCP preset BYOK: let users bring their own API key for a preset server
# [Feature] Hosted MCP preset BYOK: let users bring their own API key for a preset server

## Context (BrainPilot Cloud / hosted deployments)

In the multi-user hosted layer (`brainpilot-cloud`), we inject a set of **preset MCP servers** (e.g. `tavily`) into every user sandbox from `{usersRoot}/shared/preset_mcp_servers.json`. These presets are **read-only** for users — they cannot be edited or deleted. This is intentional: the platform owns the URL/transport/timeout configuration.

However, some preset servers (notably Tavily) are metered by an API key. We want to provide a shared key as a fallback while **encouraging users to register their own key** so that usage is billed to them, not to the platform.

## What we built on the hosted side

- Preset entries can declare a BYOK annotation:

```json
{
  "mcpServers": {
    "tavily": {
      "type": "http",
      "url": "https://mcp.tavily.com/mcp/?tavilyApiKey=PLATFORM_SHARED_KEY",
      "byok": { "kind": "tavily", "keyParam": "tavilyApiKey" }
    }
  }
}
```

- Hosted backend exposes:
  - `GET /api/mcp-servers/byok` → `[{ kind, presetName, configured }]`
  - `PUT /api/mcp-servers/byok/:kind` → body `{ apiKey }`
  - `DELETE /api/mcp-servers/byok/:kind`
- The hosted layer rewrites the preset URL with the user key when generating `bp_template/mcp_servers.json`, so the runtime only sees the final URL.
- The `byok` field is stripped before it reaches the sandbox.

## What we need from the open-source side

To keep the UX in `/app` Settings → MCP consistent, `@brainpilot/web` should understand a **preset-BYOK** mode.

### Proposed contract change

`McpServerEntry` (protocol / backend-core / web) gains an optional `byok` field:

```ts
interface McpByokInfo {
  kind: string;
  keyParam?: string;   // query parameter name to inject the key into
  keyHeader?: string;  // or header name, mutually exclusive with keyParam
}
```

When an entry has `byok`, the Settings UI should:
1. Show "Use your own key" switch / input instead of exposing the raw URL.
2. Call the hosted BYOK endpoint (`/api/mcp-servers/byok/:kind`) to save/clear the user key.
3. Still display the preset as read-only (no Delete / no manual URL edit) when the user is **not** using their own key.

On single-user / self-hosted builds (no hosted BYOK endpoint), the field can be ignored and the preset behaves exactly as today.

### Suggested minimal change

1. `packages/protocol/src/domain.ts`: add optional `byok?: McpByokInfo` to `McpServerEntrySchema` / `McpServerConfigSchema`.
2. `packages/backend-core/src/app.ts`: preserve the field in `readMcpServers` (it already reads the disk file; unknown fields can pass through or be explicitly typed).
3. `packages/web/src/components/settings/SettingsDialog.tsx` (or wherever MCP config lives): render a BYOK card for entries with `byok` when the backend advertises support.

### Detection of hosted BYOK support

Hosted deployments already expose `/api/mcp-servers/byok`. The web UI can probe it with a `GET` on settings open; if it returns 200 with an array, show BYOK inputs for matching kinds. If it 404s (single-user), fall back to current behavior.

## Why not just let users edit the preset URL?

We explicitly want to keep preset servers **read-only** in their transport shape. Letting users edit the full URL would allow them to change endpoints/headers, defeating the "managed preset" model. BYOK exposes only the credential field.

## Why not put this only in the cloud account page?

We already have a BYOK form in the closed-source `/account` page, but users naturally look for MCP configuration inside `/app` Settings → MCP. Having two places is confusing; the canonical UX should be in `/app` where the rest of the MCP servers live.

## Acceptance criteria

- [ ] `McpServerEntrySchema` accepts an optional `byok` object.
- [ ] `@brainpilot/web` Settings → MCP shows a "Bring your own Tavily key" input when `preset-tavily` (or any preset with `byok.kind === "tavily"`) is present and the backend supports `/api/mcp-servers/byok`.
- [ ] Saving the key calls `PUT /api/mcp-servers/byok/tavily`; removing calls `DELETE /api/mcp-servers/byok/tavily`.
- [ ] Self-hosted builds without the endpoint continue to work unchanged.

## Related

- Cloud-side implementation: `brainpilot-cloud` `src/mcp/{models,preset,byok,routes}.ts`.
- Open-source contract doc: `web-platform/08-open-source-contract.md` (could add this as R-15 once accepted).

Fxde42:
Closes #377.

Hosted deployments inject preset MCP servers (e.g. `tavily`) that are metered by an API key. The platform ships a shared fallback key and wants users to register their own so usage bills to them. The canonical place for that UX is `/app` Settings → MCP, where the rest of the MCP servers live — so the open-source side needs to understand the preset contract.

## Two orthogonal annotations

| field | meaning |
|---|---|
| `readOnly` | platform-managed — no Edit / Delete, and never render the raw URL (a hosted preset URL can carry the platform's shared key) |
| `byok` | metered by a credential — offer a "bring your own key" card that exposes exactly one field, never the whole transport |

A preset can be read-only without being metered, so these stay independent.

## Changes

| layer | change |
|---|---|
| `protocol/domain.ts` | `McpByokInfoSchema` (`kind` + mutually-exclusive `keyParam`/`keyHeader`); optional `readOnly` / `byok` on `McpServerEntrySchema` |
| `backend-core` | `isReadOnlyMcpServer()`; PUT/DELETE now **403** on a managed preset. `readMcpServers` already passed unknown fields through, so GET needed no change |
| `web/contracts` | `normalizeMcpServer` was a field whitelist and silently dropped both annotations — this was the real break in the chain. `serializeMcpConfig` now strips them explicitly |
| `web/utils/api.ts` | `api.mcpByok.support/save/clear` |
| `web/settings` | new `McpByokCard.tsx` + pure `mcpPresetView.ts` |
| i18n / CSS / mock | zh + en strings; card styles; a tavily preset in the mock backend so the UI is reachable in demo mode |

## Two decisions worth a look

**1. `readOnly` is enforced server-side, not just hidden in the UI.** Hosted injects presets into the same `mcp_servers.json` the user's own entries live in, so hiding the buttons is a hint — a user could still `curl` the endpoint and rewrite the preset's URL, which is exactly what the "managed preset" model is meant to prevent.

**2. `byok` / `readOnly` are on `McpServerEntrySchema` only — deliberately NOT on `McpServerConfigSchema`.** This differs from the issue's suggested minimal change. A *config* is client input; accepting these fields there would let a user POST a `readOnly: true` entry to lock themselves out of their own Delete, or fake a BYOK card. The discriminated union strips both, so the annotations can only ever be platform-injected. Covered by a test.

## Self-hosted stays unchanged

`api.mcpByok.support()` resolves to `null` (never throws) on 404, a non-JSON 2xx (the SPA `index.html` fallback serves unknown `/api` paths on some self-hosted setups), a non-array body, or a network error. `null` → no BYOK cards, and any preset renders read-only with nothing to act on. Builds without the endpoint behave exactly as they do today.

## Shared-key exposure — partially addressed

A managed entry's subtitle renders **host only**, never the query string, with a test asserting `PLATFORM_SHARED_KEY` never reaches the output. This is frontend defense only: `GET /api/mcp-servers` still returns the full URL with the shared key in it. Masking that has to happen cloud-side.

The BYOK card's key is cleared from component state on a successful save and never round-trips back — `configured` is the only state the UI reads.

## Drive-by fix

`removeMcpServer` had no try/catch. With the new 403 it would have become an unhandled rejection leaving the row silently in place; it now surfaces in the error banner.

## Testing

- `npx vitest run` — 821 passed (72 files)
- `packages/web` — 427 passed (50 files)
- `npm run typecheck`, `tsc -b`, `vite build` — all clean

42 new tests: protocol accept/reject incl. the config-strip guarantee; backend 403 + on-disk file unchanged + presets don't block CRUD on ordinary neighbours; `support()` fallback matrix; view gating and the no-shared-key-in-subtitle guarantee; zh/en string parity (`translate` falls back to the raw key, so a missing string is invisible in the UI but caught here).

Row-rendering logic lives in a pure `resolveMcpEntryView` helper rather than inline JSX because this package has no DOM harness (vitest runs in the `node` environment).


GTC2333 (dismissed):
Blocking security gap: the UI no longer renders a managed preset's full URL, but `GET /api/mcp-servers` still returns the raw stored spec via `readMcpServers()`. For the stated Tavily shape, that sends `tavilyApiKey=PLATFORM_SHARED_KEY` to every browser opening Settings, where it remains visible in DevTools/network logs.

The PR calls this out as a cloud-side follow-up, but the new feature's purpose is specifically to keep platform-managed transport/credentials out of the user-visible surface. Please make the GET representation redact the sensitive URL (while preserving enough metadata for the host-only subtitle/BYOK card), or land the required hosted-side redaction as an atomic prerequisite before merging this UI.

Fxde42:
已处理 blocking review 并同步最新 main：

- `GET /api/mcp-servers` 对 `readOnly` preset 改为严格白名单响应，仅保留安全展示元数据与 endpoint origin；完整 URL/query/path、headers、env、command/args 不再进入浏览器，磁盘配置保持不变供 runtime 使用。
- BYOK 元数据通过协议 schema 校验并移除未知字段。
- 前端 `normalizeMcpByok` 现在会 trim 注入字段，并拒绝同时提供 `keyParam`/`keyHeader` 的非法元数据。
- 新增 query/header/env/args/无效 URL 泄露回归测试。

本地验证：protocol 54/54、backend 196/196、web 454/454；root typecheck 与 web production build 均通过。请重新 review。

GTC2333 (approved):
Approved. The managed MCP response now uses a strict allowlist at the HTTP boundary, so transport credentials no longer reach the browser. The follow-up normalization and UI edge-case fixes are also covered by tests.

PR #390, opened 2026-07-29
fix: replace agent mailbox routing with a durable task ledger

### LINKED ISSUE #355: [Bug] Engineer Agent outputs Markdown blueprint spec instead of executable code, misaligned with Experimentalist wet-lab design
## Problem

When running BrainPilot multi-agent pipeline for a SAE biomarker research task, the **Engineer Agent** produced a 62KB Markdown document containing embedded R/Python code snippets, but **no independently runnable .py or .R scripts**.

More critically, the Engineer output is **completely unrelated** to what the Experimentalist designed:

### Experimentalist (correctly)
designed a **wet-lab experimental plan** with four hypotheses:
- H1: Clinical cohort N=150 ICU patients, serum S100A8/A9 ELISA + neuropsychological battery
- H2: CLP mouse model + RAGE knockout genetics, microglial NLRP3 activation readouts
- H3: HBMEC co-culture BBB model, TEER, tight junction degradation
- H4: NETs amplification loop with PAD4 knockout mice and DNase I intervention

### Engineer (wrongly)
designed a **multi-omics integration pipeline** for a hypothetical RNA-seq/proteomics/metabolomics experiment that does not exist:
- MOFA+ multi-omics factor analysis
- mixOmics DIABLO supervised integration
- WGCNA co-expression networks
- CIBERSORTx cell deconvolution
- CMap/LINCS drug repurposing
- DGIdb target identification

None of this addresses how to **compute-validate the four hypotheses using existing public GEO datasets** (e.g., GSE54514, GSE66099, GSE4607, GSE185263).

## Expected Behavior

The Engineer Agent should:
1. **Read and understand the Experimentalist hypothesis list** and design computational analyses that directly test each hypothesis
2. **Use available public datasets** (GEO series matrices, etc.) rather than assuming future multi-omics experiments will be performed
3. **Output independent, runnable script files** (.py / .R) that can be executed directly, not code snippets embedded in a Markdown document
4. Each script should produce concrete outputs: figures, tables, statistical results

## Concrete Example of What Should Have Been Produced

Instead of a generic "multi-omics blueprint", the Engineer should have written:
- h1_biomarker_validation.py — Load GEO sepsis datasets, extract S100A8/A9 expression, compute ROC/AUC vs clinical outcomes
- h2_rage_nlrp3_gsea.py — GSEA/ssGSEA of RAGE/NLRP3/NF-kB pathway genes in brain tissue transcriptomes
- h3_bbb_tight_junction.py — Differential expression of claudin-5/occludin/ZO-1 in sepsis brain; MMP-9 correlation analysis
- h4_nets_neutrophil_infiltration.py — Immune infiltration analysis, NETosis gene module vs S100A8/A9 co-expression

## Root Cause Hypothesis

The Engineer persona/prompt likely treats "Engineer" as a **computational experiment designer** rather than a **bioinformatics pipeline developer**. It needs to be instructed to:
1. Output executable scripts, not specification documents
2. Ground its analysis in real, available public data
3. Directly map each computational deliverable to a specific hypothesis from the Experimentalist

## Environment

- BrainPilot version: local copy
- Model provider: agnes-2.0-flash via custom provider
- API endpoint: http://127.0.0.1:9001
- Task domain: biomedical/bioinformatics (SAE biomarkers)

## Deliverables Actually Produced

```
deliverables/
- sae_s100a8_a9_research_report.md          (43KB, Librarian, mechanism review)
- analysis_plan_S100A8_A9_SAE.md             (26KB, Experimentalist, wet-lab protocol)
- computational_blueprint_S100A8_A9_SAE.md   (62KB, Engineer, markdown spec only - NO .py/.R files)
```

GTC2333:
## Summary

- replace `send_message`, mutable delegator routing, and mailbox delivery with persistent task records
- add `dispatch_task` and `complete_task` with stable task IDs and creator-bound reply routing
- inject compact pending task lists and one-shot completion events into agent context
- keep handoff and delegation prompt sections canonical across persona overrides
- make reminders event-driven, durable, and limited to pending assigned tasks
- pause delivery on Stop or terminal internal-agent errors without deleting unconsumed replies
- validate persisted ledgers strictly and preserve corrupt files for recovery
- emit AG-UI `task_state` snapshot and incremental events

## Review feedback addressed

- late child results route through each task own persisted `created_by`, never an agent-level mutable delegator
- the former `waitingOnPeer` heuristic is removed; reminder state is derived from explicit task operations
- stale `Handoffs`, `Delegation`, and expert communication sections are replaced before canonical contracts are injected

## Compatibility

Existing mailbox files are intentionally not migrated or read. Historical tool calls remain displayable.

## Validation

- `npm test`: 75 files, 852 tests passed
- runtime build passed
- repository TypeScript check passed
- `git diff --check` passed

Closes #355


HaoxuanLiTHUAI (changes requested):
Requesting changes after reproducing the routing behavior on this PR's head (`b0183310`). Existing CI is green, but it does not cover the interleavings below.

### [P1] Keep reply routing attached to the originating task

`entry.delegators` stores only one mutable delegator per agent. If task A delegates a child task, then task B reaches the same agent before A's child result returns, task B overwrites the map. The late result for A is subsequently rendered with task B's `reply_to`.

I reproduced this through `SessionManager`: Principal delegated task A to Engineer, Engineer delegated a child to Writer, Experimentalist then delegated task B to Engineer, and Writer finally returned A's result. The resulting envelope was:

```xml
<source type="agent" name="writer" />
<type>result_deliver</type>
<reply_to>experimentalist</reply_to>
```

The expected `reply_to` was `principal`. A runtime restart has the same underlying risk because this routing map is not persisted. Please carry task/reply correlation in the mailbox message (or otherwise serialize outstanding parent tasks) rather than deriving every result from the latest agent-level map.

### [P2] Only mark actual downstream tasks as waiting on a peer

`trace-reminder.ts` currently sets `waitingOnPeer` for every expert message whose target differs from the current delegator. A nested expert sending to `principal` is classified by `deriveMsgType` as `result_deliver`, but is still marked as waiting. Consequently `needReply` is false and neither a reply reminder nor the host fallback reaches the real delegator.

A focused regression expected a `reply` reminder and received none. This should be gated on `msgType === "task_delegate"`, not only `to !== currentDelegator`. The Auditor prompt's explicit `send_message(to="principal", ...)` should also use `<reply_to>` if nested audit delegation is supported.

### [P2 / policy clarification] Refresh non-overridable coordination sections

`appendSectionOnce` treats the presence of `## Handoffs` or `## Delegation` as proof that the canonical protocol is present. A stale/custom override can therefore retain contradictory content and suppress the new artifact and completion-criteria requirements. A regression with `## Handoffs\n\nDo not save files.` keeps that text and never injects the canonical handoff contract.

If these sections are intended to be core/non-overridable, refresh them in the same way as the communication section and add stale-override coverage. If users are intentionally allowed to replace them, please clarify that policy and adjust the "core coordination contracts survive overrides" claim.

The first issue is a merge blocker because it silently cross-routes results between independent tasks. Please add an end-to-end regression where a new task arrives before an older child result.

GTC2333:
Addressed the review in 3e19619 by replacing the mailbox/delegator model with a durable flat task ledger.

1. **Reply correlation:** every task persists its own `created_by`; `complete_task(task_id, reply)` routes completion to that creator. There is no mutable agent-level delegator. Regression coverage includes multiple creators assigning the same agent and out-of-order completion.
2. **Reminder classification:** `waitingOnPeer` and `send_message` were removed. Reminders now observe explicit `dispatch_task` / `complete_task` operations and pending assigned task IDs.
3. **Canonical coordination sections:** stale `Communicating with other agents`, `Handoffs`, and principal `Delegation` sections are removed before the canonical versions are injected. Override and idempotency tests cover this behavior.

The updated PR also retains unconsumed notifications across Stop, prevents terminal internal-agent delivery loops, and fails closed without overwriting a corrupt task ledger.

Validation: 75 test files / 848 tests passed, runtime build passed, repository typecheck passed, and `git diff --check` passed.

packages/runtime/src/extensions/trace-reminder.ts
```diff
@@ -72,191 +40,180 @@ function endedInError(e: AgentEndLike): boolean {
...
+        return;
       }
-      if (needTrace) {
-        traceRemindCount = 0;
+
+      const needTrace = state.traceWorthy && !state.traced;
+      const needReply =
+        (deps.hasPendingTasks?.() ?? false) &&
+        !state.replied &&
+        !state.dispatched;
```
HaoxuanLiTHUAI, line 180: [P1] Keep reminders task-aware when one batched task is completed

`state.replied` is run-scoped, so one successful `complete_task` suppresses `needReply` even when `hasPendingTasks()` still reports another assignment. With two assignments in the same delivery batch, completing the first produces only the trace reminder; `runDeliveryLoop` then acknowledges both assignment notifications, leaving the second task `pending` with no event that can wake the assignee. I reproduced this with two pending tasks: the reminder kind is `trace` instead of the required task-action reminder. Please base this decision on the remaining pending task set, or track acted-on task IDs, rather than a single run-wide reply flag.

packages/runtime/src/session-manager.ts
```diff
@@ -2253,50 +2308,53 @@ export class SessionManager {
...
+      if (agent.status === "error") {
+        if (agent.role === "expert") {
+          if (await this.handleDeliveryError(entry, agent)) continue;
+          await entry.taskLedger.acknowledge(notifications.map((item) => item.id));
+          return;
+        }
+        if (await this.handleInternalDeliveryError(entry, agent)) continue;
+        return;
       }
+      await entry.taskLedger.acknowledge(notifications.map((item) => item.id));
```
HaoxuanLiTHUAI: [P1] Do not acknowledge an in-flight batch after Stop aborts it

`MasAgent.abort()` normalizes an interrupted run back to `idle`, so a whole-session Stop falls through this clean-run branch and acknowledges the batch that was being processed. The task itself remains `pending`, but its only assignment notification is gone; `resumeTaskDelivery()` wakes only `notificationTargets()`, so the assignee is never resumed. A focused Stop-during-delivery test deterministically leaves the task pending while the notification count drops from 1 to 0. Preserve the batch when the run outcome was aborted, or otherwise distinguish abort from a clean provider run, so the next explicit user turn can replay it.

packages/runtime/src/session-manager.ts
```diff
@@ -2253,50 +2308,53 @@ export class SessionManager {
...
       // the sole owner of error recovery). Transient errors self-retry up to a
       // cap; fatal errors (auth/config) and the exhausted cap escalate to the
       // principal. A clean run resets the agent's consecutive-error count.
-      if (agent.status === "error" && agent.role === "expert") {
-        if (this.handleDeliveryError(entry, agent)) continue; // self-retry queued
-        return; // escalated — nothing more to drain for this agent
+      if (agent.status === "error") {
+        if (agent.role === "expert") {
+          if (await this.handleDeliveryError(entry, agent)) continue;
+          await entry.taskLedger.acknowledge(notifications.map((item) => item.id));
```
HaoxuanLiTHUAI: [P2] Preserve completed-result events on terminal delivery failure

This acknowledges every notification after a fatal or exhausted expert failure, but `writeErrorToTaskCreators()` only examines tasks still pending and assigned to that expert. If this batch contains a `completed` child-result event addressed to its creator, the child task is already terminal, so no fallback carries that result and this acknowledgement deletes its only model-visible copy. Completed tasks are omitted from `<task_list>` and there is no model tool to retrieve the ledger record after recovery. I reproduced the result queue dropping from 1 to 0 on a fatal provider error. Please retain or pause this event for replay, or explicitly forward the completed result before acknowledging it.

HaoxuanLiTHUAI (changes requested):
The previous three findings are addressed, but the task-ledger rewrite still has three reproducible delivery gaps that can strand pending work or discard a completed child result. I reproduced each case with a focused regression test.

packages/runtime/src/extensions/trace-reminder.ts
GTC2333: Fixed in 0e9677d. Reminder eligibility now reads the authoritative remaining pending-task set after tool execution; a run-level complete_task flag no longer suppresses reminders for other assignments. Added regressions for partial completion, final completion, and follow-up completion.

packages/runtime/src/session-manager.ts
GTC2333: Fixed in 0e9677d. MasAgent now records an explicit ok/error/aborted outcome. The delivery loop never acknowledges an aborted batch, and pauses delivery so the exact batch replays on the next explicit user turn. The Stop regression now asserts that both the in-flight and queued notifications remain intact.

packages/runtime/src/session-manager.ts
GTC2333: Fixed in 0e9677d. Terminal expert failure now pauses the agent and retains the entire batch instead of acknowledging it. A regression injects a completed child result, forces a fatal creator delivery failure, verifies the result remains queued, resumes delivery, and verifies the original result is replayed and then acknowledged only after a clean run.

packages/runtime/src/session-manager.ts
```diff
@@ -2253,50 +2308,70 @@ export class SessionManager {
...
-      // task_delegate in the batch). Self-retry nudges are msgType "system", so
-      // they never overwrite a real delegator recorded on the original task.
-      const delegated = [...msgs].reverse().find((m) => m.msgType === "task_delegate");
-      if (delegated) entry.delegators.set(name, delegated.fromAgent);
       this.touch(entry);
       // Surface the delegated run immediately (derived active flag, agent list).
       this.emitSessionState(entry);
       // #167: cap concurrent provider calls across experts in this session.
-      await this.withProviderSlot(sessionId, () => agent.prompt(this.renderEnvelopes(msgs, name)));
+      await this.withProviderSlot(sessionId, () =>
```
HaoxuanLiTHUAI: [P1] Recheck delivery pause after acquiring the provider slot

The pause check at the top of `runDeliveryLoop` happens before this potentially long semaphore wait. When the provider cap is full, a whole-session Stop pauses the ledger and aborts the currently running agents, but an agent waiting in `sem.acquire()` has no active prompt for `abort()` to cancel. Once a running agent releases its slot, this callback starts `agent.prompt()` without rechecking the pause; it can then complete cleanly and acknowledge a task that Stop was supposed to retain. I reproduced this deterministically with `maxConcurrentAgents: 1`: agent A occupied the slot, agent B queued here, `interrupt()` returned, and only then B was prompted. Please recheck `taskLedger.isPaused(name)` after slot acquisition before starting the provider call, or make queued acquisition cancellable; the acknowledgement path should likewise remain fenced by Stop.

HaoxuanLiTHUAI (changes requested):
The three previously reported delivery gaps are fixed in 0e9677d. One Stop race remains when task delivery is queued behind the per-session provider semaphore.

packages/runtime/src/session-manager.ts
GTC2333: Fixed in 0598ca4. Delivery now rechecks the durable pause inside the acquired provider slot before starting prompt, then uses an atomic acknowledgeIfActive ledger mutation so pauseDelivery and acknowledgement have a defined order. The deterministic maxConcurrentAgents=1 regression holds the principal slot, queues an engineer delivery, calls whole-session Stop, verifies no engineer prompt starts and the notification remains, then verifies the next explicit user message resumes and consumes it.

HaoxuanLiTHUAI (approved):

PR #394, opened 2026-08-03
feat(plugins): add marketplace control plane

GTC2333:
Stack 2/5. Depends on the plugin SDK PR.\n\nAdds disk-backed marketplace sources, immutable releases, install/enable/disable/update/rollback/uninstall lifecycle APIs, SHA-256 verification, dependency/conflict checks, and local or HTTPS registry support.\n\nValidation: backend 204 tests pass in the complete stack; focused control-plane tests cover real update and rollback.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10050,6 +10050,7 @@
       "version": "0.1.2",
       "license": "AGPL-3.0-only",
       "dependencies": {
+        "@brainpilot/plugin-sdk": "^0.1.2",
         "@brainpilot/protocol": "^0.1.2",
         "@brainpilot/runtime": "^0.1.2",
         "@hono/node-server": "^2.0.10",
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -37,6 +37,7 @@
     "start": "node ./dist/server.js"
   },
   "dependencies": {
+    "@brainpilot/plugin-sdk": "^0.1.2",
     "@brainpilot/protocol": "^0.1.2",
     "@brainpilot/runtime": "^0.1.2",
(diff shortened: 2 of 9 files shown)


Fxde42:
Review findings for commit 94943cb:

1. **[P1] Reusing an existing version directory breaks immutable-release correctness.** Reproduction: install v1, update to v2, roll back to v1, then let the catalogue serve different manifest/assets and SHA under the same v2 version and update again. `downloadRelease` verifies the new bytes, but `installBundle` returns immediately because the old v2 directory exists. The registry is then updated to the new manifest while the old assets continue to run. Please reject same-version content collisions or verify the existing installed content against the downloaded artifact before reusing it.

2. **[P2, Standards]** Compatibility logic is independently implemented in backend `compatibilityFor` and SDK `evaluatePluginCompatibility`. Their semantics already differ for undeclared engines, disabled dependencies, and bidirectional conflicts, so CLI conformance can disagree with real installation/activation. Please share one evaluator and parameterize host policy.

Fxde42 (commented):
P1 review summary:

The immutable-release invariant can be violated when a version directory already exists. `installBundle()` verifies the newly downloaded bundle, but then reuses the existing same-version directory without proving that its stored manifest/assets match the verified hash. The registry can therefore be updated to the new manifest while runtime still serves different old or tampered files from disk. A v1 → v2 → rollback-to-v1 → update-to-v2 sequence exposes this if the retained v2 directory differs from the new v2 bundle. Before reusing an immutable release, verify its persisted content hash/manifest against the incoming bundle; otherwise fail or install into a content-addressed directory.

PR #396, opened 2026-08-03
feat(plugins): add range-backed preview host

GTC2333:
Stack 4/5. Depends on the marketplace UI PR.\n\nAdds the first executable contribution host: sandboxed Preview RPC iframe, opaque file handles, bounded HTTP Range reads, dataset companion matching, enabled previewer discovery and immutable plugin assets. Ships an official Neuroscience Data Viewer for uncompressed NIfTI-1 central-slice preview.\n\nValidation: real marketplace install and enable, plugin asset load, NIfTI upload, and 206 Range smoke test passed. Runtime, backend, SDK and Web suites pass.

### DIFF
--- a/packages/backend-core/package.json
+++ b/packages/backend-core/package.json
@@ -10,7 +10,8 @@
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "files": [
-    "dist"
+    "dist",
+    "plugins"
   ],
   "repository": {
     "type": "git",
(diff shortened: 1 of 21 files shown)


Fxde42:
Review finding for commit 25ac4e2:

**[P2] Dataset companion discovery depends on the selected file suffix being included in `companions`.** In `openRangePreview`, a rule such as `companions: [".eeg", ".vmrk"]` opened from `x.vhdr` produces no `selectedSuffix`, so no companions are discovered at all. If `.vhdr` is included, opening `x.VHDR` reconstructs `x.vhdr`, which fails on a case-sensitive filesystem. Please derive the stem from the matched previewer extension/current basename and preserve the selected file path exactly; also validate or handle required suffixes independently of the companions list.

PR #397, opened 2026-08-03
feat(runtime): add isolated subagent workers

GTC2333:
Stack 5/5. Depends on the preview-host PR but is isolated from GoT v2.\n\nAdds bounded leaf-agent batches with fresh contexts, isolated histories and workspaces, validated profiles, structured submit_result, explicit input manifests, artifact publication, background wait/get/cancel tools, persisted lifecycle state, Agents-panel visibility and per-child usage accounting. The first child may borrow the paused parent provider lease to avoid concurrency-one deadlock.\n\nExisting installs receive new editable profile templates through an idempotent lightweight scaffold upgrade. session_state.subagents remains optional for older clients.\n\nValidation: Runtime 498, Backend 204, Web 467, CLI 107 and SDK 8 tests pass; production Web build and live mock subagent smoke test pass.


Fxde42:
Review findings for commit 0925279:

1. **[P1] Reference inputs are not read-only or workspace-isolated.** `data/shared` inputs default to reference mode and expose a real absolute path. The `code-runner` profile has `bash`, while the managed-path guard only rewrites `/workspace` for bash and does not block writes elsewhere. A child can therefore run `printf x > <manifest path>` and mutate the parent persistent library. This violates the isolated scratch workspace/read-only input contract. Enforce filesystem isolation for bash or materialize read-only copies/mounts.

2. **[P1] Background children can bypass provider concurrency.** With max concurrency 1, `spawn_subagent(wait=false)` launches a child that inherits and borrows the parent lease, but the parent is no longer paused after the tool returns and proceeds to its next provider request. Parent and child can then issue two requests under one slot, causing 429s and violating the stated lease rule. Borrow only for the synchronous wait path, or transfer/reacquire the lease before the parent resumes.

3. **[P3, non-blocking]** The five subagent lifecycle tools are repeated across role allowlists in `AGENT_TOOL_CONFIG`; a shared allowlist would reduce permission drift.

Fxde42 (commented):
P1 review summary:

1. Reference-mode inputs break child isolation. `materializeInputs()` gives a subagent the real absolute path of parent persistent data, while the child still has unrestricted `bash`. A code-runner can write directly to the manifest path and mutate the parent library. Reference inputs must be mounted/read through an enforced read-only boundary, copied into scratch, or withheld from roles with writable shell access.

2. `spawn_subagent(wait=false)` can bypass the per-session provider concurrency cap. The child borrows the paused parent's provider lease, but the parent immediately continues and can start another provider request while the child is still using that lease. With a limit of one this creates two concurrent provider calls and can trigger 429s. Lease borrowing must remain valid only while the parent is actually suspended, or background children must acquire their own provider slot before running.

GTC2333:
Resolved the latest-main conflicts in 539c79d.

The two textual conflicts were integrated as capability unions:
- SessionEntry now owns both SubagentManager and the GoT checkpoint/audit state.
- ToolDeps and role allowlists retain isolated subagent lifecycle tools together with GoT v2 readers/review tools. Auditor remains unable to record or edit ordinary Trace nodes, while retaining bounded review access and permitted leaf subagents.

Validation:
- full repository suite passed: 82 files / 908 tests
- full workspace production build passed, including Web and docs
- subagent manager, integration, profile, GoT, checkpoint, plugin, and merged tool-access tests all pass

#397 is mergeable again; CI is running and approval is required.

Fxde42 (approved):

PR #399, opened 2026-08-03
feat(got): add workspace checkpoints and causal restore

GTC2333:
## Summary

Adds the checkpoint and irreversible restore layer on top of GoT v2.

- captures per-session Git-backed workspace snapshots when record_trace runs
- attaches checkpoint and file provenance to Trace nodes
- provides bounded diff, restore preview, restore, causal rollback preview, and causal rollback APIs
- rolls back only Auditor-confirmed causal descendants
- blocks restore while agents, tools, user input, or delivery loops are active
- materializes and validates both target and recovery trees before changing workspace files
- applies the target and creates its post-restore baseline before cancelling pending tasks and clearing the delivery list
- restores the original workspace and preserves queued work if target application, baseline creation, or queue persistence fails
- keeps a session locked if both the restore and automatic recovery fail, so old work cannot run against an uncertain workspace
- uses internal recovery checkpoints only for transaction recovery
- adds Git to the shared runtime baseline used by the default CPU sandbox

## Restore semantics

Restore and causal rollback are intentionally irreversible after success. Completed task history remains available; pending tasks are then cancelled and all queued notifications are cleared. A failed restore preserves the old workspace and queue whenever automatic recovery succeeds. Recovery checkpoint IDs are not exposed and no Undo operation is provided.

## Deliberately excluded

- Web UI
- Undo
- Principal GoT context injection
- plugins, subagents, and neuroscience preview changes

## Validation

- typecheck passed
- full runtime suite passed: 51 files / 508 tests
- focused checkpoint and TaskLedger suite passed: 21 tests
- regressions cover corrupt target materialization, callback persistence failure, automatic recovery, quota-saturated restore, conflicts, and post-restore baselines

## Stack

- Depends on #401, which follows merged #398
- Followed by #400

### DIFF
--- a/docker/main/Dockerfile
+++ b/docker/main/Dockerfile
@@ -39,7 +39,7 @@ RUN if [ -n "$APT_MIRROR" ]; then \
       sed -i "s|http://deb.debian.org|$APT_MIRROR|g" /etc/apt/sources.list.d/*.sources 2>/dev/null || true; \
     fi \
     && apt-get update \
-    && apt-get install -y --no-install-recommends curl ca-certificates \
+    && apt-get install -y --no-install-recommends curl ca-certificates git \
     && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /app
--- a/docker/sandbox/Dockerfile
+++ b/docker/sandbox/Dockerfile
@@ -43,8 +43,8 @@ RUN if [ -n "$NPM_REGISTRY" ]; then npm config set registry "$NPM_REGISTRY"; fi 
 # (protocol ← runtime), not the whole monorepo (web/cli/backend-core).
 RUN npx tsc -b packages/runtime
 
-# ---- python-baseline: node + python3 + pip + curl + 镜像源 ----
-#   cpu 链的根，也是 gpu-base 的父（python+curl 层只构建一次，被两条链复用）。
+# ---- python-baseline: node + python3 + pip + curl + Git + 镜像源 ----
+#   cpu 链的根，也是 gpu-base 的父（共享运行时依赖只构建一次，被两条链复用）。
 FROM node:22-slim AS python-baseline
 ARG APT_MIRROR=""
 ARG PIP_INDEX_URL=""
--- a/docker/sandbox/Dockerfile
+++ b/docker/sandbox/Dockerfile
@@ -75,6 +75,8 @@ CMD ["node", "packages/runtime/dist/server.js"]
 #   不在此安装 → monorepo 代码变更只重建下面的业务层，base 层完全复用。
 #   ⚠️ base tag 写死（经典构建器不支持 FROM ${ARG}）；升级须同步三处（见文件头）。
 FROM ghcr.io/neuroaihub/brainpilot-gpu-base:cu124-torch2.6.0 AS gpu
+# Older published GPU base tags may predate workspace checkpoints.
+RUN command -v git >/dev/null 2>&1 || (apt-get update && apt-get install -y --no-install-recommends git && rm -rf /var/lib/apt/lists/*)
 WORKDIR /app
 COPY --from=builder /app/node_modules ./node_modules
 COPY --from=builder /app/packages ./packages
--- a/docker/sandbox/_python-base.sh
+++ b/docker/sandbox/_python-base.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
-# _python-base.sh — cpu/gpu sandbox 共享的 python3 + pip + curl + 镜像源安装段。
+# _python-base.sh — cpu/gpu sandbox 共享的 python3 + pip + curl + Git + 镜像源安装段。
 # 由 Dockerfile 的 python-baseline stage 直接 RUN（gpu-base 继承该 stage，无需重跑）。
-# curl/ca-certificates 供 HEALTHCHECK 探活用，cpu/gpu 都需要，故装在此共享层。
+# curl/ca-certificates 供 HEALTHCHECK 探活；Git 供 GoT workspace checkpoint
+# 捕获、diff 与恢复使用。它们都是 cpu/gpu 运行时依赖，故装在此共享层。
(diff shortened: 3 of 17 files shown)


Fxde42:
Review findings (commit `aba00bb`):

1. **[P1] The default CPU sandbox still has no Git.** Git is added to `docker/main/Dockerfile` and the GPU stage, but `cpu` inherits `python-baseline`, whose script installs only Python/pip/curl/CA certs. The normal `--target cpu` runtime therefore makes every checkpoint unavailable. Please install Git in the shared baseline or CPU stage.

2. **[P1] Paused queued work bypasses the restore gate.** `hasAnyAgentActivity()` ignores notification targets when their agent is paused. Such an agent can retain durable queued tasks; restore is allowed, then the queued work can later continue against a rewritten workspace. The PR explicitly requires blocking restore while queued tasks exist.

3. **[P1] Restore leaves the checkpoint base on the pre-restore tree.** `restore()` / `restoreCausal()` capture a recovery checkpoint (which becomes `headCheckpointId`) and then apply the target without moving the head to the resulting workspace. The next Trace checkpoint attributes the whole restore to the next node; rolling that node back can reintroduce changes intentionally restored away. Record/set the actual post-restore head before accepting later captures.

GTC2333:
Addressed finding 1 in a6ad067: Git is now installed in the shared python-baseline inherited by the default CPU sandbox, and the image build runs git --version so a missing runtime dependency fails the build. The existing GPU compatibility fallback remains in place. Local shell syntax and diff checks pass; this environment cannot access the Docker daemon, so the actual image build is left to CI. Findings 2 and 3 are not yet resolved and remain under discussion.

GTC2333:
Addressed the two remaining findings in `5ed5fe3` following the agreed destructive-rollback semantics:

2. A workspace operation still blocks all live agents/tools/delivery loops. Durable queued work no longer bypasses into the restored workspace: after a successful restore, every pending task is marked cancelled and the complete notification/delivery queue is cleared before the workspace lock is released. Completed task history remains intact.
3. Both full and causal restore now create an internal post-restore `baseline` checkpoint and make it the new head. The pre-restore recovery checkpoint is retained only for transactional failure recovery; if baseline creation fails, the workspace is restored to that recovery state and the operation fails. The next Trace checkpoint therefore diffs from the actual restored tree.

Undo support was intentionally removed from the public protocol and runtime transaction log. Regressions cover queue cancellation and verify that the next checkpoint contains only edits made after either restore type. Full non-Web suite: 77 files / 877 tests passed; typecheck and production builds pass.

GTC2333:
Additional failure-hardening in `83c6672`:

- Queue cancellation now runs through an `onPrepared` hook only after the target, state token, causal conflicts, Git objects, and pre-restore recovery checkpoint have all been validated successfully. Predictable restore failures therefore leave pending work untouched.
- Once prepared, pending tasks are durably cancelled before any workspace file is changed, so a later infrastructure failure cannot release old work against a partially/restored workspace.
- Internal recovery and post-restore baseline checkpoints bypass the ordinary Trace-history quota; a full checkpoint repository no longer prevents restoring an existing checkpoint.
- Failed checkpoint capture restores the previous head instead of leaving it pointed at an incomplete record.

Targeted regressions cover stale/conflicting preflight hooks, quota-saturated restore, and correct post-restore baselines. Typecheck and 19 focused tests pass; no additional deployment test was performed.

Fxde42:
Re-review of `4159af5` + `5ed5fe3`:

- Fixed: Git is now installed in the CPU sandbox, and the post-restore baseline prevents the next checkpoint from inheriting the whole restore diff.
- **[P1] The queued-work/Undo requirements were replaced with destructive behavior rather than fixed.** The PR body says restore is blocked while queued tasks exist and recovery checkpoints are retained for Undo. The new gate deliberately ignores all queued notifications, mutates the workspace first, then `cancelAllPending()` cancels tasks and clears the entire delivery queue; recovery IDs are removed from the API. This can discard completed-result/system notifications as well as pending work, and Undo is impossible. Please reject before workspace mutation when any queue is non-empty and retain the recovery transaction/Undo contract. If irreversibility is a new product decision, the PR spec needs explicit approval and must be rewritten before this implementation can be reviewed against it.

GTC2333:
Updated the PR specification to the approved irreversible-rollback contract.

The current head 83c6672 also supersedes the mutation order described in the review: after all predictable preflight checks and recovery capture succeed, pending tasks are durably cancelled and the complete delivery list is cleared before any workspace file is changed. Completed task history remains available. Internal recovery checkpoints are failure-transaction safeguards only; public Undo and recovery IDs are deliberately excluded.

Fxde42 (commented):
P1 review summary:

Rollback preparation is still not failure-safe. The code durably cancels pending tasks and clears the complete delivery queue in `onPrepared`, but only afterwards materializes/checks out the target tree in `applyRecord()` / `applyTree()`. A missing or corrupt Git object, checkout failure, or equivalent target-materialization error returns a failed restore with the original workspace recovered or untouched, while queued work has already been irreversibly destroyed. Fully materialize and validate the target in temporary storage before invoking the destructive preparation callback, then make the workspace swap the only remaining operation.

GTC2333:
Addressed in c491c6c with a stronger success-only commit boundary.

Both the target and recovery trees are fully materialized and validated before workspace mutation. The session-level workspace lock prevents agents, delivery, uploads, and new messages throughout the operation. The target is applied and its post-restore baseline is created before the TaskLedger callback cancels pending work and clears notifications.

If target application, baseline creation, or the durable TaskLedger callback fails, the pre-materialized recovery tree is applied and the checkpoint head is reset; queued work remains unchanged because TaskLedger mutations roll back on persistence failure. If automatic recovery itself fails, the error is marked WORKSPACE_RECOVERY_FAILED and SessionManager deliberately keeps the workspace locked.

Regressions cover corrupt target materialization and final callback failure. Full runtime suite: 51 files / 508 tests; focused checkpoint/ledger suite: 21 tests; typecheck passed.

PR #400, opened 2026-08-03
feat(got): add audit and checkpoint Web UI

GTC2333:
## Summary

Adds the user-facing GoT v2 experience on top of the core and checkpoint layers.

- renders canonical V2 confirmed, candidate, uncertain, and rejected parent states across initial REST loads and live trace_delta snapshots
- lays out all parent conclusions as one DAG, with a defensive recursion guard for malformed legacy/external payloads
- adds node details, checkpoint file lists, per-file diffs, restore confirmation, and causal rollback
- warns that a successful rollback is irreversible, cancels pending tasks, and clears the delivery list
- displays append-only Trace change history and Auditor reports
- consumes trace_delta snapshots without regressing to stale revisions
- keeps the existing Principal GoT view and prompt injection unchanged

## Restore semantics

Workspace restore and causal rollback are intentionally irreversible after success. The Web UI does not expose Undo or recovery checkpoint IDs.

## Deliberately excluded

- Undo
- plugin marketplace and plugin UI
- subagent UI
- neuroscience preview UI
- automatic Auditor-to-PI context injection

## Validation

- Web production build passed
- Web full suite passed: 54 files / 464 tests
- focused normalization coverage uses the actual materialized GET producer shape and all four canonical parent conclusions
- layout coverage verifies all-state ranking and malformed-cycle fail-closed behavior

## Stack

- Depends on #401 and #399


Fxde42:
Review findings (commit `1a31d51`):

1. **[P1] Real V2 live deltas lose every causal/candidate edge.** Runtime nodes always contain `parents: [{ nodeId, conclusion }]`. `normalizeTraceGraph()` sees an array and calls `normalizeTraceNode()`, which only reads legacy `{ id }`; canonical parents are filtered out and dependency reconstruction is skipped. The test uses producer-unrealistic nodes with no `parents`. Normalize `nodeId/conclusion` and add a producer-shaped regression test.

2. **[P1] Undo can desynchronize workspace and GoT state.** After Undo, the component stores the newly returned recovery checkpoint as another Undo target. A second click flips the workspace back, but the backend restores revoked statuses only for the recovery ID recorded on the original `causal_rollback`; workspace contents and node state diverge. Undo must be single-use or transactionally restore both states each time.

3. **[P2] Undo disappears after refresh/reopen.** The UI only retains the recovery ID in component state or reads `node.metadata.recoveryCheckpointId`, while the backend stores it only in change metadata. Rehydrate the latest applicable recovery transaction from persisted history.

GTC2333:
Addressed findings 1 and 3 in `1230f44`:

1. Canonical runtime parents `{ nodeId, conclusion, reason }` are normalized directly into the UI edge shape; rejected parents are omitted. The regression now uses a producer-shaped V2 snapshot and verifies both confirmed and candidate edges.
3. The checkpoint detail reloads persisted Trace changes and reconstructs the latest causal rollback for the selected node that has not been consumed by a matching `causal_rollback_undo`. This also works for branch points without their own checkpoint.

Semantic-link UI and adapters were removed consistently with the core protocol decision. Finding 2 (repeat Undo semantics) remains intentionally open for the design discussion.

Validation: Web build passed, 54 files / 463 Web tests passed, and the stacked full suite passed 77 files / 876 tests plus the all-workspace production build.

GTC2333:
Design update for finding 2/3 in `e39a0ea`: BrainPilot no longer exposes Undo for workspace or causal rollback. Rollback is explicitly irreversible. The recovery checkpoint remains backend-internal only for reverting a failed transaction; its ID is no longer returned by the protocol, stored in Trace change metadata, rehydrated by the UI, or rendered as an Undo action.

Both confirmation dialogs now warn that the operation cannot be undone and will cancel all pending tasks and clear the delivery list. The previous repeat-Undo desynchronization and refresh/reopen state therefore no longer exist.

Validation: 54 Web test files / 462 tests passed and the Web production build succeeds.

Fxde42:
Re-review of `a663ce7` + `e39a0ea`:

- Fixed partially: runtime-shaped live `trace_delta` parents now accept `{ nodeId, conclusion, reason }`.
- **[P1] Initial/refresh REST graphs still lose non-confirmed canonical parent states.** `SessionManager.getTrace()` returns `getGraph()`, whose materialized `parents` contains confirmed parents only and whose full canonical set is in `causalParents`. `normalizeTraceGraph()` sees that legacy `parents` array and returns early without consuming `causalParents` or rebuilding from top-level dependencies. Candidate/uncertain edges therefore disappear until a later live delta. In addition, `normalizeTraceNode()` explicitly filters `rejected`, so rejected state is never rendered even from live canonical data. This violates “renders canonical V2 parent states.” Add a regression test using the actual GET producer shape.
- **[P1] Undo was removed, not repaired.** The PR body still promises “causal rollback, and undo,” but the component, protocol result and labels now make every restore irreversible. Please restore a persisted single-use/transactional Undo flow rather than deleting the requirement.

GTC2333:
Addressed the remaining canonical-parent issue in afd023b and updated the PR specification to the approved irreversible-rollback contract.

Materialized GET nodes now prefer causalParents over the confirmed-only parents compatibility view, retain rejected edges, and preserve confirmed/candidate/uncertain/rejected states across refresh. The regression uses the actual SessionManager.getTrace producer shape. The four states also receive distinct graph styling.

Undo remains intentionally excluded: rollback is irreversible and the confirmation explains pending-task cancellation and delivery-list clearing.

Validation: 54 Web test files / 463 tests passed, and the Web production build passed.

Fxde42 (commented):
P1 review summary:

Rendering all canonical parent conclusions introduces a valid graph-layout crash. Core cycle checks ignore rejected edges, so history may legitimately contain A → B rejected followed by B → A candidate. `normalizeTraceNode()` now retains both states, while `buildTraceLayout()` recursively ranks through every parent without a visited/in-progress guard. Opening that graph causes `RangeError: Maximum call stack size exceeded`. Rank using only the confirmed DAG, or add cycle-safe ranking for non-confirmed edges while still drawing candidate/uncertain/rejected history.

GTC2333:
Addressed according to the clarified product invariant: every canonical parent conclusion participates in one DAG, not only confirmed edges.

Follow-up #401 enforces this in the Host for candidate, confirmed, uncertain, and rejected parents, rejects reverse proposals through rejected history, and repairs malformed persisted cycles by dropping the later cycle-closing edge. The Web layout therefore continues to rank using all displayed parent states. Commit 01a300d adds an in-progress guard only as fail-closed protection for malformed legacy/external payloads, so such input cannot crash the panel.

Validation: 54 Web test files / 464 tests and the Web production build passed; the stacked runtime suite passed 51 files / 508 tests.

PR #401, opened 2026-08-03
fix(got): enforce a DAG across all parent states

GTC2333:
## Summary

Follow-up to merged #398 that applies the accepted graph invariant to every canonical parent conclusion.

- treats candidate, confirmed, uncertain, and rejected parent edges as one DAG
- rejects a new candidate whenever it would close a cycle through any existing parent state
- repairs malformed persisted cycles by dropping the later cycle-closing edge
- retains all parent states in the API/UI graph; state changes do not change topology
- makes neighborhood traversal set-based and defensively normalizes record_trace artifact arrays

## Deliberately deferred

The separate review finding about Trace re-proposals changing an already confirmed parent is not changed in this PR, per product decision.

## Validation

- runtime typecheck passed
- focused Trace core tests passed: 16 tests
- the stacked runtime suite passed after rebasing: 51 files / 508 tests


Fxde42 (dismissed):

GTC2333:
Resolved the latest-main conflict in 791f85e.

The only textual conflict was the backend forward helper: the merged version keeps GoT checkpoint routes with named path params and plugin preview routes with Range-header forwarding. All other GoT/plugin overlaps merged automatically.

Validation after installing the merged lockfile dependencies:
- full repository test suite: 79 files / 889 tests passed
- full workspace production build passed, including Web and docs
- targeted plugin control-plane and checkpoint HTTP tests passed: 32 tests

#401 is mergeable again. The branch update dismissed the previous approval, so it now only needs CI completion and renewed approval.

Fxde42 (approved):

PR #436, opened 2026-08-07
fix(runtime): bound trace checkpoints and resolve data paths

### LINKED ISSUE #433: fix(runtime): resolve physical roots before logical /data aliases
## Problem

When `BP_DATA_DIR=/data`, the agent workspace and persistent library use physical paths under the same prefix that is also treated as a logical alias:

- workspace: `/data/workspaces/<session>`
- persistent library: `/data/data`
- logical `/data/...`: persistent library

Structured file tools currently interpret every `/data/...` as a logical path before checking whether it is already a real path under the current workspace or persistent directory. This rewrites workspace paths to `/data/data/workspaces/...` and can rewrite `/data/data/...` to `/data/data/data/...`. Bash still uses the physical path, so Bash and read/write/edit disagree about the same string.

## Root cause

`rewriteLogicalPath` selects `/data` by string prefix before classifying known physical roots. Existing tests use `/root/.bp-root`, so they do not cover the production collision.

## Proposed fix

- For absolute paths, classify current workspace, attachments, persistent, and shared physical roots first.
- Apply logical aliases only when the path is not already under a known physical root.
- Keep HTTP Files API logical-path behavior unchanged.
- Add regression tests with `cwd=/data/workspaces/s1` and `persistentDir=/data/data`.

## Acceptance criteria

- Structured tools and Bash address the same workspace file.
- Real persistent paths are not rewritten twice.
- Logical `/data/x` still maps to `/data/data/x`.
- `/workspace/x` remains session-scoped.
- Existing traversal and cross-session confinement tests continue to pass.

### LINKED ISSUE #434: fix(trace): bound checkpoint provenance and exclude generated environments
## Problem

A workspace-local virtual environment can make one `record_trace` invocation serialize tens of thousands of checkpoint changes into the Trace Agent prompt. The same file list is later registered into the Trace artifact graph.

Observed consequences include million-character Trace inputs, `stopReason=length`, large checkpoint repositories, many Git provenance lookups, and graph pollution by dependency files.

## Root cause

- Checkpoints hard-exclude only `.attachments` and `.truncated`.
- Snapshot limits are byte-based but have no file-count bound.
- `record_trace` injects the full `Git-Evidence` array into the Trace prompt.
- Checkpoint attachment registers every changed file as an artifact.
- The first Trace checkpoint may diff against the empty tree, attributing pre-existing workspace files to the first Trace event.

## Proposed fix

Initial safe patch:

- Exclude common generated environment/cache roots from checkpoint capture.
- Add a snapshot file-count cap.
- Replace full prompt provenance with a bounded checkpoint summary and sample.
- Bound provenance used for automatic artifact registration while retaining the checkpoint reference and on-demand detail APIs.
- Add stress tests using a large fake venv.

Follow-up design:

- Capture a baseline before the first agent run so Trace checkpoints represent agent changes rather than the initial workspace.
- Prefer explicit `artifact_outputs` for graph registration.

## Acceptance criteria

- A workspace containing a large venv produces a bounded Trace envelope.
- Generated dependency files do not enter checkpoint provenance or the artifact graph.
- Trace Agent can complete without context-length termination.
- Checkpoint diff/restore for ordinary workspace artifacts still works.
- Truncation is explicit and deterministic.

GTC2333:
## Summary

- resolve absolute paths under known workspace/persistent roots before applying the logical `/data` alias
- add the `/data`-mounted production regression case
- establish an initial checkpoint baseline only when pre-existing workspace inputs exist
- exclude generated environments and caches from checkpoint capture
- cap checkpoint file count and provenance expansion
- replace full Trace prompt provenance with a bounded summary and sample

## Why

This fixes two failure chains observed in Bench runs:

1. Bash and structured file tools could resolve the same `/data/workspaces/...` path to different locations.
2. A workspace-local virtual environment could be serialized into Trace prompts and artifact provenance, producing extreme token and storage amplification.

## Validation

- Runtime TypeScript build/typecheck passes.
- Targeted regression suite: 102 tests passed.
- Full runtime suite: 569 tests passed.
- `git diff --check` passes.

Closes #433
Closes #434

### DIFF
--- a/packages/runtime/src/__tests__/managed-path-rewrite.test.ts
+++ b/packages/runtime/src/__tests__/managed-path-rewrite.test.ts
@@ -109,6 +109,42 @@ describe("rewriteLogicalPath", () => {
     expect(r.path).toBe(abs);
   });
 
+  it("prefers physical roots when the data root itself is mounted at /data", () => {
+    const mountedCwd = resolve("/data/workspaces/s1");
+    const mountedPersistentDir = resolve("/data/data");
+    const mountedRoots: ManagedPathRoots = {
+      cwd: mountedCwd,
+      persistentDir: mountedPersistentDir,
+    };
+
+    const workspace = rewriteLogicalPath("/data/workspaces/s1/scripts/train.py", mountedRoots);
+    expect(workspace).toMatchObject({
+      ok: true,
+      path: "/data/workspaces/s1/scripts/train.py",
+      rewritten: false,
+      root: "workspace",
+      abs: resolve(mountedCwd, "scripts/train.py"),
+    });
+
+    const physicalData = rewriteLogicalPath("/data/data/models/best.pkl", mountedRoots);
+    expect(physicalData).toMatchObject({
+      ok: true,
+      path: "/data/data/models/best.pkl",
+      rewritten: false,
+      root: "data",
+      abs: resolve(mountedPersistentDir, "models/best.pkl"),
+    });
+
+    const logicalData = rewriteLogicalPath("/data/models/best.pkl", mountedRoots);
+    expect(logicalData).toMatchObject({
+      ok: true,
+      path: resolve(mountedPersistentDir, "models/best.pkl"),
+      rewritten: true,
+      root: "data",
+      abs: resolve(mountedPersistentDir, "models/best.pkl"),
+    });
+  });
+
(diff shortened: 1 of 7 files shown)


GTC2333:
Closing because commit 602bf6c is already contained in upstream dev; no additional merge is needed.

PR #449, opened 2026-08-11
feat(runtime): complete stage-one harness and research workflows

GTC2333:
## Summary

This is the cumulative Stage 1 PR targeting `main`. It includes the approved foundation from #448 plus the task/subagent harness, research workflow, and related Web agent-network integration previously reviewed in #450 and #451.

- inject durable task notifications into active agent runs instead of racing a second prompt
- distinguish task replies from completion and expose aggregate session work state
- support shared-workspace research subagents with bounded artifacts and capacity states
- preserve plugin-independent aggregate work tracking
- surface shared subagent lifecycle in the Web agent network

## History repair

After #448 was squash-merged, this branch still contained #448's original commits, which caused a history-only merge conflict. The branch was rebased onto the current `main`, dropping those duplicate commits. The resulting file tree is identical to the pre-rebase PR head; no product code changed during this repair.

## Verification

- `npm run typecheck`
- full test suite: 1002/1003 passed; the sole failure was a transient concurrent temp-directory cleanup error
- isolated rerun of the affected `session-manager.test.ts`: 50/50 passed

## Review context

- supersedes the stacked review branches in #450 and #451
- ready for final approval once required checks pass


Fxde42 (commented):
P2 — Settled execution promises are never released

`launch()` stores every execution promise in `this.executions`, but settled promises are never removed. Because each promise retains its task, context, profile, and result, sessions that launch many subagents will accumulate memory for their entire lifetime.

Could we remove the entry in a `finally` block, guarded by an identity check so a newer execution cannot be deleted accidentally? Completed lookups can continue using `this.results`.

GTC2333:
Addressed the review feedback in commit 8306a92: settled subagent execution promises are removed in finally, guarded by promise identity so an older execution cannot delete a newer one. Completed results remain available through waitFor. Added a regression test covering cleanup and result retrieval.

Fxde42 (dismissed):

GTC2333:
Merge conflict repaired by rebasing onto the squash-merged `main` and removing only the duplicate #448 history. The rebased head has the exact same file tree as the previous PR head, so this repair introduces no content changes.

Local validation: typecheck passed; full suite passed 1002/1003 with one transient concurrent temp-directory cleanup failure, and the affected session-manager suite passed 50/50 in isolation.

Fxde42 (changes requested):
Requesting changes for three concrete user-impacting issues:

1. **P1 — Ordinary Principal requests trigger an unnecessary extra model turn and a false workflow warning.**

   SessionManager always installs the Principal workflow guard, while renderPrincipalWorkflowState only checks whether a qualifying delegation exists. It never determines whether the current request is substantive scientific execution. As a result, even a simple question reaches the agent_end handler with a non-empty state, queues a follow-up through sendUserMessage, and Pi starts another provider turn. If no delegation occurs, the next agent_end emits the misleading research workflow remains incomplete warning. This adds latency and token cost to ordinary conversations and can produce duplicate output. Please keep the guard and reminder inactive for exempt requests, not merely describe the exemption in prompt text.

   Relevant code: packages/runtime/src/session-manager.ts:2471-2479, 2586-2609; packages/runtime/src/extensions/principal-workflow-guard.ts:92-107.

2. **P1 — Background subagents bypass the provider concurrency limit.**

   startBatch launches wait:false children and returns immediately, so the parent resumes. However, prepareBatch still gives the first child borrowParent=true. The child inherits the provider async context and withProviderSlot takes the borrowed lane instead of the global semaphore. With maxConcurrentAgents=1, parent and child provider requests can therefore overlap, defeating the documented bound and causing rate-limit failures. Please allow lease borrowing only for the synchronous waiting path; background batches must acquire normal provider capacity.

   Relevant code: packages/runtime/src/subagent-manager.ts:176-185, 284-298; packages/runtime/src/session-manager.ts:2640-2650.

3. **P2 — Restore and rollback can run while a shared-workspace child is active.**

   deriveWorkActive includes queued, waiting, and running subagents, but hasAnyAgentActivity, which gates restore and rollback, does not. A wait:false child using the default shared workspace can therefore continue writing while the backend restores or replaces that workspace, corrupting the restored state or losing child output. The Web button normally blocks this, but the REST API and a state change after confirmation remain exposed. Please include active subagent states in the workspace-operation gate.

   Relevant code: packages/runtime/src/session-manager.ts:2954-2985, 3101-3157.

I intentionally excluded branch metadata, documentation bookkeeping, task display ordering, the workspace-dot input case, timeout micro-races, and general code smells because they do not materially affect the core user experience.

_AI-assisted review, narrowed and verified against the repository scope guidance._

Fxde42 (approved):

PR #450, opened 2026-08-11
feat(runtime): enforce research orchestration policy

GTC2333:
## Summary

- require PI delegation for substantive scientific execution
- add Engineer environment, accelerator, and end-to-end preflight gates
- broaden method selection before deep evaluation
- require representative real-data iteration and evidence handoffs
- add the host-owned Principal workflow guard
- preserve shared-workspace research handoff policy

## Verification

- 116 targeted runtime/protocol/client-cli tests
- protocol, runtime, and client-cli typecheck

## Dependency

Stage 1 PR 3/4, stacked on #449. Merge after #448 and #449, then retarget this PR to `main`.


Fxde42 (changes requested):
P1 — The delegation guard applies to every Principal request

The host-owned guard currently activates whenever no qualifying Expert task exists after the epoch baseline. Since the baseline is reset for every idle Principal message, simple questions, summaries, and document-only requests are also forced to delegate.

This exceeds the PR stated scope of “substantive scientific execution” and conflicts with the exemptions in `personas.ts`.

Could we gate this mechanism on an explicit substantive-research classification, leaving it inactive for exempt requests?

GTC2333:
Addressed the review concern while preserving the bounded agent_end follow-up behavior. The context guidance and single automatic follow-up are now explicitly conditional on substantive scientific execution, use advisory wording, and state that simple questions, document-only work, and summaries of already-validated results do not require delegation. No request classifier was added. Unit and SessionManager integration assertions cover the revised contract.

PR #451, opened 2026-08-11
feat(research): add bounded auditor and research skills

GTC2333:
## Summary

- decouple GoT provenance from scientific auditing
- add specialist evidence, data-integrity, code-artifact, and method-validation audit skills
- bound parallel review scope, retries, and evidence chronology checks
- add the official research plugin with source-grounded reporting and data inventory skills
- refine structured skill search and non-English technical query guidance
- retain empirical iteration history and compact actionable audit reports

## Verification

- 167 targeted runtime and plugin tests
- full repository typecheck
- full repository test run: 998/1002 passed initially; the remaining 3 Azure patch tests and 1 parallel temp-directory cleanup test all passed on isolated rerun after executing the required postinstall patch

## Dependency

Stage 1 PR 4/4, stacked on #450. Merge after #448, #449, and #450, then retarget this PR to `main`.


Fxde42 (commented):
P2 — `keyword_hits` no longer represents keyword hits

The weighted relevance `score` is assigned to both `keyword_hits` and `score`. Previously, `keyword_hits` represented an occurrence count, while the new value includes field weights such as six points for a name match. Existing consumers may therefore receive a relevance score where they expect a hit count.

Could we preserve `keyword_hits` as an actual occurrence count, or remove/deprecate it as an intentional API change?

GTC2333:
Addressed the review feedback in commit abaaa61: keyword_hits again reports the legacy case-insensitive occurrence count in the skill description, while the new field-weighted value remains available separately as score and continues to drive ranking. The regression test now asserts keyword_hits equals 1 and score equals 13 for the fixture.

PR #453, opened 2026-08-12
fix(runtime): normalize optional tool schemas

### LINKED ISSUE #452: [Bug]: Kimi intermittently rejects all-optional tool schemas across multiple agents
## Bug Description

The Kimi OpenAI-compatible route intermittently rejects requests **before inference** when an exposed tool has an object-parameter schema with no required fields.

Representative response:

```text
HTTP 400
Tool 6 function has invalid 'parameters' schema:
None is not of type 'array'

On schema['required']:
None
```

The source schemas omit `required`, which is valid JSON Schema. The Kimi compatibility/validation layer appears to materialize the missing value as `required: null` and then rejects it. Other providers accept the same tool schemas.

This is not specific to one agent or one tool.

## Production Evidence

Window: 2026-07-17 through 2026-08-12.

- 1,705 Kimi assistant/model calls
- 78 total failed calls
- 36 failures with the exact invalid-`required` schema error
- Since 2026-08-11: 100 Kimi calls, including 36 schema rejections
- Rejected calls report zero token usage, confirming rejection before inference

Observed affected agents and tool mappings:

| Agent | Kimi calls | Total errors | Schema errors | Reported tool | Resolved tool |
|---|---:|---:|---:|---|---|
| experimentalist | 107 | 33 | 16 | Tool 6 | `ls` |
| principal | 573 | 24 | 11 | Tool 6 | `ls` |
| librarian | 239 | 7 | 6 | Tool 7 | `search_papers_local` |
| trace | 476 | 13 | 3 | Tool 4 | `get_trace_graph` |

The affected schemas all have no required parameters:

- `ls`: optional `path` and `limit`
- `search_papers_local`: all search filters optional
- `get_trace_graph`: empty object input

`engineer` and `writer` also expose `ls`; `auditor` exposes `search_papers_local`. They have not produced this schema error in the incident window, but they are structurally exposed to the same failure.

The issue is intermittent: successful Kimi requests and schema rejections occur close together with unchanged agent configuration, suggesting inconsistent validation behavior across gateway/backend nodes.

## Steps to Reproduce

1. Configure a BrainPilot session to use the Kimi OpenAI-compatible provider.
2. Expose at least one tool whose object schema has no `required` field, for example:

   ```json
   {
     "type": "object",
     "properties": {
       "path": { "type": "string" },
       "limit": { "type": "number" }
     }
   }
   ```

3. Start an agent turn.
4. Observe intermittent HTTP 400 responses reporting that `required` is `None` instead of an array.

Existing BrainPilot paths that reproduce the shape include `principal`/`experimentalist` with `ls`, `librarian` with `search_papers_local`, and `trace` with `get_trace_graph`.

## Expected Behavior

A tool schema with no required fields should be accepted, or serialized as the semantically equivalent `required: []`. All BrainPilot agents using Kimi should be able to start inference normally.

## Actual Behavior

Kimi intermittently rejects the entire request with HTTP 400 before inference. Repeated orchestration/delivery attempts can turn one compatibility failure into multiple failed agent runs and inflate the per-agent failure rate.

## Deployment

- Docker deployment
- BrainPilot repository version: `0.1.0`
- Investigated revision: `d53b2379`
- Runtime Node.js: v22.23.1

## Root Cause Assessment

BrainPilot/the current `pi-ai` OpenAI-compatible conversion passes `tool.parameters` through without normalization. The Kimi route does not consistently tolerate an omitted `required` field and reports it as `null`/`None`.

The external gateway behavior is the immediate trigger, but BrainPilot can prevent the failure defensively at the provider-request boundary.

### P0: Normalize all tool schemas at the final provider boundary

Before serializing an OpenAI-compatible request, recursively normalize object schemas:

```ts
function normalizeToolSchema(schema: unknown): unknown {
  // Deep-clone and recursively visit properties/items/oneOf/anyOf/allOf.
  // For every object schema:
  //   - preserve an existing string[] required value
  //   - replace missing/null/non-array required with []
}
```

Then serialize:

```ts
parameters: normalizeToolSchema(tool.parameters)
```

This should happen after built-in, system, extension, and MCP tools have been combined. Fixing only `ls` or only `adaptTool()` is insufficient because:

- built-in tools such as `ls` bypass the system-tool adapter;
- custom tools such as `get_trace_graph` and `search_papers_local` also reproduce the issue;
- MCP tools may introduce the same valid all-optional schema shape.

Preferred long-term location: the `@earendil-works/pi-ai` OpenAI-compatible `convertTools` path.

Safe BrainPilot hotfix: a `before_provider_request` hook that normalizes every `payload.tools[*].function.parameters` entry for Kimi/OpenAI-compatible providers.

Adding `required: []` is semantically equivalent to omitting `required`, so applying this normalization to all OpenAI-compatible providers should be low risk.

### P1: Prevent failure amplification

- Classify this exact schema-validation 400 separately from transient provider 400s.
- Do not repeatedly launch identical agent runs with the same invalid payload.
- Surface one actionable error or fail over to another configured provider.
- Log the resolved tool name as well as the provider's numeric tool index.

## Acceptance Criteria

- [ ] Serialized schemas for `ls`, `search_papers_local`, and `get_trace_graph` contain `required: []`.
- [ ] Existing non-empty `required` arrays remain unchanged.
- [ ] Nested object schemas are normalized recursively.
- [ ] Built-in, system, extension, and MCP tools all pass through the same normalization.
- [ ] Unit tests cover missing `required`, `required: null`, empty objects, nested objects, and existing required fields.
- [ ] A Kimi provider smoke/contract test no longer returns this HTTP 400.
- [ ] Other OpenAI-compatible providers continue to receive semantically equivalent schemas.
- [ ] Deterministic schema failures do not trigger an unbounded sequence of agent runs.

Fxde42:
## Summary

- normalize object tool schemas at the final OpenAI Chat Completions provider boundary
- cover built-in, custom, MCP, and active extension tools after Pi combines them
- preserve existing required arrays and leave other provider APIs unchanged
- add unit coverage and an end-to-end Kimi-compatible contract regression test

Closes #452

## Testing

- `BP_MOCK=1 npm test -w @brainpilot/runtime` — 573 passed
- `npm run typecheck`
- `BP_MOCK=1 npx vitest run packages/runtime/src/__tests__/tool-schema-compat.test.ts packages/runtime/src/__tests__/openai-tool-schema-compat.test.ts` — 7 passed

## AI assistance

This PR was implemented and self-reviewed with AI assistance.


HaoxuanLiTHUAI (approved):
Reviewed after syncing current main. The conflict resolution preserves both the principal workflow guard and the OpenAI tool-schema compatibility extension, with the compatibility normalization kept at the final provider boundary. Local typecheck, 1015 non-web tests, 504 web tests, docs checks/build, web build, and full workspace build all pass; GitHub CI #664 also passes.

Issue #462, opened 2026-08-17
release: prepare the post-0.1.2 npm prerelease and close release gates


HaoxuanLiTHUAI:
## Summary

Current main is substantially ahead of npm `latest@0.1.2`, but it is not publish-ready yet.

## Current evidence

- All workspace package versions are still `0.1.2`.
- `npm publish --workspaces --if-present --dry-run --access public` fails because published packages cannot overwrite 0.1.2.
- `CHANGELOG.md` has no post-0.1.2 entry.
- New packages such as `@brainpilot/plugin-sdk`, `@brainpilot/plugin-auditor`, `@brainpilot/plugin-got`, and `@brainpilot/plugin-research` are not yet present on npm.
- `npm audit --omit=dev` reports 12 production findings: 7 high and 5 moderate.
- Local docs build needs the optional native Tailwind oxide package when npm omits it.

Main is 56 commits / 296 files ahead of v0.1.2 and includes task-ledger/subagent changes, Graph of Trace v2, checkpoints, plugin SDK/marketplace, datasets, and new system plugins.

## Proposed release strategy

Use a prerelease rather than updating `latest` immediately. Because the change includes intentional runtime/state migrations and new public packages, `0.2.0-beta.0` is the preferred candidate.

## Acceptance criteria

- Fix the runtime/UI/sandbox release-blocking defects tracked separately.
- Triage all production audit findings and document any accepted residual risk.
- Choose and synchronize the prerelease version across every publishable workspace.
- Add a complete changelog and upgrade notes.
- Verify first-publish access/provenance for new scoped packages.
- Run `npm ci`, typecheck, tests, build, pack inspection, and publish dry-run in a clean release environment.
- Publish with the `beta` dist-tag, soak on the 208 test environment, then decide whether to promote to `latest`.

HaoxuanLiTHUAI:
Maintainer release decision update: the target is now the stable `v0.2.0` release, not `0.2.0-beta.0`. Release freeze has started on exact `main` `35b5aada`. This issue remains useful for the security/package/permission gates, but its prerelease recommendation is superseded by the current release plan and #491 coordination. No publishing action has started.

PR #464, opened 2026-08-17
fix: close stream, provider, and sandbox release blockers

### LINKED ISSUE #460: fix(sandbox): provide executable ripgrep and file utilities in CPU/GPU images
## Summary

Real GPU sandbox runs complete, but the built-in grep tool can fail or stall because `rg` is unavailable or not executable. The sandbox also lacks the `file` utility.

## Evidence

On main `3732f9e3`, the EEG alpha QC end-to-end case completed but recorded:

- `ripgrep (rg) is not available and could not be downloaded` after ~121 seconds
- `spawn /tmp/home/.pi/agent/bin/rg EACCES` after ~77 seconds
- another `rg EACCES` failure from the auditor
- a shell command exited 127 when it reached `file figures/eeg_alpha_qc.png`

The case recovered through other tools, but the failures added several minutes and make normal research workflows unreliable.

## Root cause

The shared sandbox baseline installs Python, curl, CA certificates, and Git, but not `ripgrep` or `file`. The Pi fallback download is not reliable in the deployed sandbox.

## Acceptance criteria

- Install `ripgrep` and `file` in the shared CPU/GPU sandbox baseline.
- Fail image builds if `rg --version` or `file --version` cannot execute.
- Add/extend Docker smoke coverage for both image targets.
- Confirm the built-in grep tool no longer downloads a private fallback binary.

### LINKED ISSUE #461: fix(backend): resolve apiKeyEnv when testing Environment provider profiles
## Summary

Settings → Providers → Test reports an env-bootstrapped provider as `unavailable` even though real conversations through the same provider succeed.

## Reproduction

On main `3732f9e3`, the active profile is:

- name: Environment
- API: anthropic-messages
- base URL: `https://api.deepseek.com/anthropic`
- model: `deepseek-v4-flash`
- key source: `env:$ANTHROPIC_API_KEY`

A real prompt completes successfully. Clicking **Test** returns `unavailable` in ~358 ms and persists an HTTP 401 health result.

## Root cause

Env-bootstrapped profiles intentionally persist `apiKey: ""` plus `apiKeyEnv` so secrets are not written to disk. The provider test route passes `p.apiKey` directly to `probeProvider` and does not resolve `process.env[p.apiKeyEnv]`.

## Expected behavior

The probe should use the same effective credential resolution semantics as runtime requests without persisting or returning the plaintext key.

## Acceptance criteria

- Resolve `apiKeyEnv` at probe time when the stored key is empty.
- Keep the secret out of `providers.json`, logs, API responses, and error messages.
- Preserve explicit stored-key behavior.
- Add route tests for successful env-backed probes and missing env variables.
- Ensure the persisted health result reflects the real probe.

### LINKED ISSUE #463: fix(web): preserve repeated stream deltas emitted in the same millisecond
## Summary

Conversation hydration can silently delete legitimate model output when two identical stream deltas share the same millisecond timestamp.

The web reducer currently fingerprints append events as `type + stream id + _ts + delta`. The runtime emits timestamps with millisecond precision, so a model that emits the same token twice within one millisecond produces an identity collision. The second token is treated as a history/SSE replay duplicate.

## Reproduction

Deployed main `3732f9e3` on the 208 test environment and asked for:

```
Answer briefly what is 2+2. Show the equation in LaTeX.
```

Persisted events contain:

```
2 + 2 = 4.

$$2 + 2 = 4$$
```

After `reduceMessagesForEvent`, the UI content becomes:

```
2 + 2 =4.

$$2 +  =4$$
```

A longer multi-agent run also corrupted paths such as `/data/workspaces/<id>/...` and deleted characters from scientific text.

## Root cause

`packages/web/src/contexts/messageReducer.ts` uses `_ts + delta` as the stable replay key, while `packages/runtime/src/events.ts` only supplies millisecond-resolution `_ts` and no unique event sequence.

## Expected behavior

- Distinct emitted events must never collapse, even when timestamp and delta are identical.
- The same persisted event replayed through history and SSE must still be idempotent.
- Text, reasoning, and tool-argument streams must follow the same rule.

## Acceptance criteria

- Add a transport-level stable event identity/sequence.
- Use that identity for stream replay deduplication.
- Preserve legacy fallback behavior for histories without the new identity.
- Add regression tests with repeated identical deltas sharing exactly the same `_ts`.
- Cover text, reasoning, tool args, and history + SSE replay.

HaoxuanLiTHUAI:
## Summary

- add a persisted `_event_id` to runtime events and use it for stream replay deduplication, preserving identical text, reasoning, and tool-argument deltas emitted in the same millisecond
- resolve `apiKeyEnv` (including `.env` fallback) when testing an Environment provider profile, and report a clear missing-variable error without sending an empty credential
- install and assert `ripgrep` and `file` in CPU/GPU sandbox images

## Validation

- `npm run typecheck`
- `BP_MOCK=1 npm test` — 92 files, 1018 tests
- `npm --workspace @brainpilot/web test` — 63 files, 506 tests
- `npm --workspace @brainpilot/web run build`
- `npm run version:check`
- `npm pack --workspaces --if-present --dry-run`
- built the `python-baseline` Docker target and smoke-tested `git`, `rg`, and `file` inside the image

## Remaining release gate

`npm publish --dry-run` correctly refuses to republish the existing `0.1.2`. The version bump/tag/publish work remains tracked in #462; this PR intentionally does not change package versions or publish artifacts.

Closes #460
Closes #461
Closes #463

Related: #462


Fxde42 (approved):
Standards: correct; no high-impact repository-standard violations found. Spec: the stream identity, env-backed provider probe, and CPU/GPU sandbox tool fixes satisfy the linked issues on supported paths. CI is green.

PR #471, opened 2026-08-17
fix: harden main UX workflows and feature activation

### LINKED ISSUE #465: fix(web): preserve concrete tool names in live activity summaries
## Summary

Live tool activity frequently renders as the generic label “调用工具：tool” even though the persisted `TOOL_CALL_START` includes a concrete name such as `bash`, `read`, or `record_trace`.

## Reproduction

Run a task that invokes built-in tools and observe the activity summary while the tool is active.

## Observed

- Raw events contain `tool_call_name`.
- The live activity summary displays `tool`.
- After completion the group collapses into a generic thinking-step count, making it difficult to identify what actually ran.

## Expected

The concrete, friendly tool name should be visible throughout the live lifecycle and after refresh/rehydration.

## Acceptance criteria

- Regression tests cover live SSE, session-state reconciliation, and history hydration.
- A known `tool_call_name` never degrades to the generic fallback.
- MCP names continue to use the existing friendly `server · tool` presentation.

### LINKED ISSUE #466: fix(datasets): ship and preflight downloaders used by automatic recipes
## Summary

The dataset marketplace advertises several entries as automatic public downloads, but the main runtime image does not ship the command-line tools used by their recipes.

## Reproduction

On `main@3732f9e3`, start downloads from the dataset marketplace.

## Observed on the 208 preview

- OpenNeuro fails with: `Required downloader 'datalad' is not installed or not on PATH`.
- The main container also lacks `dandi` and `wget`.
- BCI Competition IV 2a succeeds because its recipe uses the built-in HTTP downloader.
- A completed BCI job still presents “下载数据 / 开始下载” instead of a completed/open-location action.

## Expected

Every recipe advertised as an automatic download must be executable in the published main image, or the UI must preflight and disable it with actionable setup guidance before starting a doomed job.

## Acceptance criteria

- Published main images support the `datalad`, `dandi`, and `wget` recipes, with build-time smoke assertions.
- Downloader readiness is covered by tests.
- Completed jobs do not present the same primary action as a never-started job.
- Retry remains available for failed jobs.

### LINKED ISSUE #467: fix(runtime): do not treat trivial attachment reads as substantive scientific execution
## Summary

A trivial attachment read can be classified as substantive scientific execution, triggering the Principal workflow guard, unnecessary trace work, and repeated final replies.

## Reproduction

Upload a small text file named `brainpilot-safari-upload-test.txt` and ask:

> Read the attached synthetic test file and reply with exactly the value after session-isolation-marker, with no other text.

## Observed

- The Principal correctly reads the 135-byte file.
- It emits the requested marker three times as three distinct persisted assistant messages.
- A warning is persisted: `Principal ignored the required Expert-delegation reminder; the research workflow remains incomplete.`
- The trivial task consumes about 69,970 tokens including trace processing.

This is not history/SSE replay duplication: the three messages have different `message_id` values.

## Root cause

`isSubstantiveScientificExecutionRequest()` treats `attached ... file` as concrete input and matches the adjective `test` in “test file” as an input-bound execution action.

## Acceptance criteria

- Attachment notices and filenames do not influence execution-intent classification.
- “Read/quote/extract this test file” stays non-substantive.
- “Test/evaluate/analyze the attached dataset” remains substantive.
- The reproduction produces one final answer and no delegation violation.

### LINKED ISSUE #468: fix(web): derive session titles from user text instead of attachment notices
## Summary

When the first prompt contains an attachment, the generated session title starts with the internal attachment notice instead of the user's question.

## Reproduction

1. Start a draft conversation.
2. Attach any file.
3. Enter a normal question and send.

## Observed

The new session is titled:

`[本次对话附件（位于 .attachments/ 目录）：brainpilot-safari-u...`

The relevant user text is not visible in the sidebar.

## Root cause

`SessionContext.sendPrompt()` calls `createSession(trimmed.slice(0, 48), ...)`, where `trimmed` is the transport content after `PromptComposer` prepends the localized attachment notice.

## Acceptance criteria

- New session titles are derived from visible user-authored text.
- Attachment names may be a fallback only when the user supplied no text.
- Chinese and English attachment notices are excluded.
- Add a web regression test for draft → attached send → generated title.

### LINKED ISSUE #469: fix(plugins): reload MCP projections after enable/disable without runtime restart
## Summary

Enabling an installed MCP plugin does not make its tools available to subsequently created sessions until the runtime process restarts.

## Reproduction

1. Deploy `main@3732f9e3`.
2. Enable the verified **Playwright Browser Automation** plugin.
3. Create a new conversation after the enable operation.
4. Ask the Principal to use the Playwright MCP.

## Observed

- The marketplace immediately reports the plugin as enabled.
- `/data/plugins/runtime/org.brainpilot.playwright-mcp.json` exists.
- The new Principal reports that no Playwright MCP tools are exposed and falls back to `bash`/`curl`.

## Root cause

`SessionManager.ensureMcpTools()` caches `mcpLoaded` and `mcpTools` at manager scope. Once an earlier session has loaded an empty MCP configuration, later plugin projection changes are never re-read.

## Expected

New agents created after plugin enable/disable should use the latest runtime projections without restarting the runtime. Existing running agents may remain unchanged.

## Acceptance criteria

- Regression coverage for no plugin → enable projection → new session receives tools.
- Regression coverage for enabled → disable projection → new session no longer receives tools.
- Obsolete MCP clients are closed when configuration changes.
- Existing agents keep their current tool list for their lifetime.

### LINKED ISSUE #470: fix(runtime): prevent completion while delegated Expert tasks are pending
## Summary

The Principal can announce completion and emit a final-looking answer while an Expert task is still pending, then emit a second final answer after the Expert result arrives.

## Reproduction

Ask the Principal to delegate one bounded checklist task to Engineer, create a report, and summarize only after all work is complete.

## Observed

1. Principal dispatches the Engineer.
2. Principal says “All steps completed” and renders a full summary.
3. Engineer completes later.
4. Principal resumes, verifies the artifact, and renders another full completion summary.

The first summary incorrectly claims the delegated artifact is complete before its result is available.

## Expected

Pending tasks created by the Principal should prevent a completion-state response. Interim updates must be clearly marked as pending, and the user should receive exactly one final completion summary after all required task results settle.

## Acceptance criteria

- Add a deterministic task-ledger regression covering an active Principal with a pending created task.
- Prevent or steer premature completion while required tasks remain pending.
- Preserve normal interim progress messages.
- Emit one final completion response after the child result is delivered.

HaoxuanLiTHUAI:
## Summary

- keep trivial attachment reads out of the substantive-science workflow gate and avoid trace follow-ups for conservative read-only shell chains
- strengthen delegated-task context and dispatch results so the Principal waits for pending Expert results before claiming completion
- hot-reload MCP plugin projections for newly-created agents after enable/disable changes
- derive draft session titles from user-authored text and preserve concrete tool names on result activities
- ship the automatic dataset recipe prerequisites in a separately testable main-image stage and remove repeat-download actions for completed jobs

## Root causes

- attachment transport notices and the noun phrase `test file` leaked into the workflow intent classifier
- read-only `bash` calls armed the trace reminder even when they only listed or inspected files
- MCP tools were cached once at `SessionManager` scope, so later plugin projection changes were ignored
- draft titles were sliced from the transport payload after the attachment notice was prepended
- `TOOL_CALL_RESULT` created a second activity row without inheriting the originating call name
- the main image advertised DataLad/DANDI/wget recipes without shipping their executables

## Validation

- `BP_MOCK=1 npm test` — 92 files, 1028 tests
- `npm --workspace @brainpilot/web test` — 67 files, 531 tests
- `npm run typecheck`
- `npm --workspace @brainpilot/web run build`
- `npm run version:check`
- built the `runtime-deps` Docker stage and smoke-tested DataLad 1.6.0, git-annex, DANDI 0.76.7, and wget inside the image

## Docker note

The independently built downloader stage measured about 342 MB on local ARM. A full local `runtime` target also reaches the repository's pre-existing ARM-only npm optional-dependency failure for `@tailwindcss/oxide-linux-arm64-gnu`; the downloader stage itself builds and runs successfully. GitHub CI remains the authoritative full-image/build gate.

## Related work

The same Safari run also reproduced stream-delta corruption (#463), which is already fixed separately in draft PR #464 and is intentionally not duplicated here.

Closes #465
Closes #466
Closes #467
Closes #468
Closes #469
Closes #470


Fxde42 (changes requested):
Standards: no high-impact repository-standard violations found. Spec has three blocking supported-flow issues: (1) on every MCP config change, SessionManager closes the shared bridge, so tools already handed to existing agents still call clients that have just been closed; preserve usable per-generation clients for existing agents and close them only when no agent depends on them. (2) the TOOL_CALL_RESULT fix only copies a name from the current live reducer state; MESSAGES_SNAPSHOT still converts role=tool results without toolName and never reconciles them with the flattened assistant toolCalls, so refresh/history hydration still renders generic tool results. (3) intentText strips the verb from the common Chinese request 请测试附件中的数据集 because TEST_INPUT_NOUN consumes 测试附件, causing a substantive attached-dataset test to bypass the Principal workflow guard. Please add regressions for old-agent tool execution after enable/disable, snapshot tool-result naming, and this Chinese classifier path.

HaoxuanLiTHUAI:
@Fxde42 感谢 review。重新核对当前 head 后，三项 blocking feedback 都已在 `7e32149d` 及后续提交中处理：

1. MCP 配置刷新不再关闭旧 Agent 仍依赖的 client。每代工具闭包保留其 client，旧 Agent 可继续执行；`session-manager.test.ts` 覆盖了启用后创建的 Agent 在后续禁用配置后仍能调用原 MCP 工具，bridge 仅在 manager shutdown 时关闭。
2. `MESSAGES_SNAPSHOT` 现在通过 `toolCallId` 将 role=tool 结果与 assistant `toolCalls` 对齐并恢复 `toolName`；`messageStreamIdempotency.test.ts` 有刷新/历史 hydration 回归。
3. `请测试附件中的数据集` 已明确判定为 substantive scientific execution；分类器修正了“测试附件中/里/内…”路径，并加入对应中文回归测试。

分支已在 `e432408a` 合并最新 main，当前 build-and-test、blackbox-fast 和两项 CodeQL 均通过，GitHub 状态为 mergeable。烦请有空时重新 review；如果仍有未覆盖的行为路径，我再按具体复现继续处理。

Fxde42 (changes requested):
Core-impact re-review found one blocking issue:

**P2 — `packages/web/src/components/plugins/DatasetMarketplace.tsx:123`**

`listDatasetJobs()` returns jobs newest-first, but `jobsByDataset` is built with `new Map(jobs.map(...))`, so older attempts overwrite the newest attempt for the same dataset.

In the ordinary supported flow **download fails → user retries → retry completes**, the UI continues to read the older failed job. It therefore keeps showing failure, never exposes the completed state, and offers another download action.

Please preserve the first/newest job for each dataset instead of the last one, and cover failed → retry → completed with a regression test.

The other changed core workflows look correct under normal supported use.

HaoxuanLiTHUAI:
Contributor update: the latest P2 re-review blocker is addressed in `d6f5edb5`.

Root cause: `listDatasetJobs()` returns newest-first, but constructing `new Map(jobs.map(...))` let later, older attempts overwrite the latest job for the same dataset. In a failed → retry → completed flow, the UI could therefore keep rendering the stale failure and offer another download.

The Web UI now indexes the first/newest job per dataset and never lets older attempts replace it. A regression covers an older failed attempt followed by a newer completed retry, including the completed-state action guard.

Validation:
- targeted Dataset Marketplace tests: 7/7
- full Web suite: 547/547
- production Web build: passed
- GitHub build-and-test, blackbox-fast, and both CodeQL jobs: passed

The PR is mergeable at the code/check level; `CHANGES_REQUESTED` remains until re-review. Contributors touching dataset download state should preserve the API's newest-first ordering contract when building per-dataset indexes.

Fxde42 (approved):
Re-reviewed at d6f5edb. The requested dataset retry fix is correct, the regression test covers the failed-to-completed retry path, and all required CI checks pass. No remaining core-impact findings.

Issue #521, opened 2026-08-21
[Bug][UX][Safari] New conversation keeps a stale Files URL and can hide a completed first reply


HaoxuanLiTHUAI:
## Summary

On current `main` (`f7be504a`) in Safari against the isolated 208 service-parity deployment, creating a new conversation while the previous session is on a Files route changes the in-memory session but leaves the browser URL on the previous session/path.

## Reproduction

1. Open `/app`; the restored route is `/sessions/<old>/files?path=/workspace/app`.
2. Click **New conversation**.
3. Select a real provider/model and send `只回复：SAFARI_SHORT_OK`.
4. Backend creates session `eb3bb30d-...` and finishes with exact reply `SAFARI_SHORT_OK` in 2.5 s.
5. UI remains on “正在准备发送” because the URL/SSE context still points at the old Files route.
6. Clicking the new session in the sidebar immediately rehydrates the completed reply.
7. Hard-refreshing the `/sessions/.../files` URL exits the workbench to the marketing home page; navigating back through `/app` restores it.

The stale path also opens the Files panel with “链接指向的文件不存在或当前不可访问。” for `/workspace/app`.

## Expected

- Draft -> created-session promotion must atomically replace the URL with the new session route.
- SSE/history/files scopes must switch to the same session id.
- Hard refresh of an app session route must serve/recover the authenticated SPA, not the marketing page.
- The Files route must not invent `/workspace/app` from the `/app` entry path.
- Add Safari/WebKit coverage for first-send promotion, completed-before-navigation reply, and hard refresh.

HaoxuanLiTHUAI:
Implemented as two independent Draft PRs: OSS session/draft route reset in #526, and hosted hard-refresh routing in BrainPilot-cloud#111. OSS: 26 targeted + 647 full Web tests and production build passed. Cloud: frontend-serving tests, typecheck, and full 329-test suite passed. Please review these PRs instead of starting duplicate fixes.

HaoxuanLiTHUAI:
Completed by OSS PR #526 and BrainPilot-cloud PR #111. Both are merged. The combined 208 Safari pass verified fresh-session first reply, stale Files-route reset, and valid deep-link hard refresh; the final exact-main deployment is healthy at OSS `35b5aada` / Cloud `72f241d3`.

PR #526, opened 2026-08-21
fix(ux): keep new conversations out of stale Files routes

HaoxuanLiTHUAI:
## Summary

Partial OSS fix for #521; the hosted hard-refresh companion is tracked in the linked Cloud PR.

- excludes `/app` and other unambiguous application roots from legacy naked-file parsing, so `/app` is no longer converted to `/workspace/app`
- resets a canonical `/sessions/:id/files` URL back to `/app` when the user switches conversations or opens a draft
- clears the stale file request while keeping the Files panel preference
- preserves a copied file deep link until its owning session/file is resolved once

## Verification

- `npm test -w @brainpilot/web -- workspaceFileLink.test.ts sessionSelection.test.ts` — 26 passed
- `npm test -w @brainpilot/web` — 80 files / 647 tests passed
- `npm run build -w @brainpilot/web` — production build passed


HaoxuanLiTHUAI:
208 Safari combined-QA passed the real route/session cases:

- From an existing `/sessions/.../files?path=/workspace/app` URL, clicking New conversation returned to clean `/app` instead of retaining the stale Files route.
- The first real-provider prompt created the new session and rendered exact `INTEGRATED_MAIN_OK` as the first reply.
- A valid `/sessions/eb3bb30d.../files?path=/workspace/safari-note.txt` deep link retained session/file selection and previewed exact `SAFARI_FILE_OK` after hard refresh.

lucasli1015 (approved):
Reviewed the route parser and session-transition reset. Initial copied deep links are preserved until handled, later draft/session changes leave stale Files routes, application roots are no longer misclassified, and valid hosted hard-refresh behavior was verified on 208/Safari. No blocking findings.

PR #527, opened 2026-08-21
fix(runtime): keep internal reminders out of Stop recovery

### LINKED ISSUE #522: [Bug][UX] Stop/resume can render NO-RENDER control text and consume the first follow-up
## Summary

Current `main` (`f7be504a`) can expose an internal Trace reminder marker as a normal Principal message after Stop, and the first recovery prompt does not receive its requested answer.

## Reproduction (real provider, Safari, 208 QA)

1. In a working session send: `请用 bash 运行 sleep 8，结束后只回复 SLEEP_DONE。`
2. The assistant shows `SLEEP_DONE`, but the UI still says `principal 正在工作` because an internal Trace reminder continues.
3. Click **停止生成**.
4. Send `只回复：RECOVER_OK`.
5. Chat renders this as the assistant reply:
   `<!--NO-RENDER-->trace reminder handled<!--/NO-RENDER-->`
6. Send a second prompt `只回复：RECOVER_OK_2`; this one succeeds.

The same marker is visible multiple times in a long restored conversation.

## Expected

- NO-RENDER/internal system acknowledgements must never become visible chat messages.
- A user follow-up after Stop must not be consumed by an older internal reminder.
- If the visible answer is complete but internal cleanup continues, the UI should distinguish “finalizing history” from “model is still answering”, or finish cleanup without keeping the main Stop state active.
- Add regression coverage for Stop during a post-answer Trace reminder and immediate follow-up delivery.

Follow-up/regression context: #134, #489, and #502.

HaoxuanLiTHUAI:
## Summary

Closes #522.

- clears Pi steering/follow-up queues before and after abort settlement, covering the `agent_end` reminder race
- prevents an internal Trace/workflow reminder queued before Stop from consuming the next explicit user prompt
- buffers split `<!--NO-RENDER-->...<!--/NO-RENDER-->` stream chunks without ever rendering them
- removes the completed internal-only message at TEXT_MESSAGE_END
- drops atomic NO-RENDER chat chunks and keeps old persisted histories clean on rehydrate

## Verification

- Runtime targeted: `event-mapping.test.ts` + `interrupt-cancel.test.ts` — 22 passed
- Web targeted: `demoTruncatedExport.test.ts` + `messageStreamIdempotency.test.ts` + `internalToolStrip.test.ts` — 30 passed
- Runtime full: 65 files / 657 tests passed
- Web full: 80 files / 648 tests passed
- Runtime and Web production builds passed


HaoxuanLiTHUAI:
208 Safari real-provider retest found and fixed one additional Stop boundary:

- NO-RENDER control text is now hidden correctly.
- However, after Stop, the next request still saw the prior aborted user instruction in model history and executed its `sleep 12` before answering the new request. History confirmed this was a fresh model decision, not a replayed SDK tool call.
- Commit `de9488b2` adds an internal, non-rendered boundary only when resuming a whole-session Stop: unfinished instructions/tools/output from the interrupted turn are cancelled, and the model must follow only the new user message. The visible/durable chat event remains the exact user text.

Validation:
- Stop/resume targeted suite: 21 passed
- full Runtime suite: 65 files / 658 tests passed
- Runtime build and `git diff --check` passed

I am redeploying the combined QA image and will repeat the exact Safari scenario before marking this resolved.

HaoxuanLiTHUAI:
208 Safari real-provider retest passed after deploying commit `de9488b2` in the combined QA image (`a78c7eb6`):

1. New session: asked Runtime to execute `sleep 12` and return `OLD_OUTPUT`.
2. Stopped while the bash tool was running; tool result became `Command aborted` and the run ended `aborted`.
3. Sent `请只回复：NEW_ONLY_OK`.
4. Recovery returned exact `NEW_ONLY_OK` in 2.2s, with no bash/tool replay and no NO-RENDER text in Chat.
5. Persisted Pi history shows the internal interruption boundary followed by the exact new user text; the visible AG-UI user message remains only the original user text.

This closes the additional real-world Stop/resume gap found during combined QA.

lucasli1015 (approved):
Reviewed queue clearing, split NO-RENDER buffering, and the whole-session Stop boundary. Visible AG-UI user text remains unchanged while only the model input receives the cancellation marker. Full Runtime/Web tests and the exact 208/Safari Stop recovery scenario passed. No blocking findings.

PR #533, opened 2026-08-22
fix(runtime): make Stop a hard model-context boundary

### LINKED ISSUE #532: [Bug][P0][UX] Stopped turn can resume tools and Trace after the recovery prompt
## Summary

Production v0.2.0 still allows an explicitly stopped turn to continue after the next user prompt. The recovery prompt is answered, but stale tool work, Trace processing, file writes, and the old final answer resume afterward. This is a stricter regression of #522 / #527.

## Exact production reproduction

Environment: `brainpilot.chat`, Cloud v0.1.10, CPU sandbox `brainpilot-sandbox:0.2.0` digest `sha256:a627c781516d588ff7f28771369790504cdfb2037f4063c52979bc4e317bdf9b`, real provider `deepseek-v4-flash`, reasoning `high`.

1. Send: `请列出 100 条脑成像质量控制检查项，每条都写完整解释。`
2. While Principal is working, click **停止生成**.
3. UI correctly shows `用户已中断当前任务` and enables the composer.
4. Immediately send: `已停止。只回复：可以继续。`
5. Principal displays `可以继续。` after a few seconds.
6. The run incorrectly remains active for ~100 seconds.
7. A stale `write` tool and Trace task then complete the cancelled 100-item request, create `docs/reports/brain-imaging-qc-checklist.md`, and inject the old report after the recovery answer.

## Runtime evidence

- Stop emitted `RUN_FINISHED` and `runState.active=false` for the cancelled turn.
- The next explicit user message started a new run and received the visible requested answer.
- The same new Pi turn then executed stale tool/Trace context from the cancelled request.
- Final state became idle only after the obsolete report and Trace reminder completed.

## Impact

Stop is not a trustworthy cancellation boundary: an obsolete task can mutate workspace files, consume tokens, generate misleading Trace records, and overwrite the conversational intent after the user has moved on.

## Expected

- Stop must cancel all unfinished instructions, tool calls, files, and Trace work from that turn.
- The next user prompt must run from the last completed Pi branch, excluding the interrupted user/assistant/tool context.
- The visible stopped user message may remain in BrainPilot history for audit, but it must not remain on the active model-context branch.
- Add a regression test that aborts after an incomplete tool-producing turn, sends a recovery prompt, and asserts the stale tool/Trace work never executes.

## Proposed minimal direction

Capture the Pi `SessionManager` leaf before each top-level prompt. After abort fully settles, branch the active Pi session back to that checkpoint (or reset for the first turn), in addition to the existing two-sided `clearQueue()`. Then append the recovery prompt on the clean branch.

Related: #522, #527.

HaoxuanLiTHUAI:
Closes #532.

Production v0.2.0 QA reproduced a stricter Stop regression after #527: the immediate recovery prompt was answered, but the interrupted turn's stale tool/Trace work then resumed, wrote a report, and injected the old answer.

This patch captures the Pi SessionManager leaf before each top-level BrainPilot prompt. Once Stop has fully settled, it branches the persisted Pi session and the live in-memory Agent message state back to that completed checkpoint. BrainPilot's visible audit history remains unchanged; only unfinished model/tool context is excluded from the next prompt. Existing two-sided queue clearing remains in place.

## Validation so far

- targeted Stop/event mapping: 24 passed
- runtime full suite: 65 files / 671 tests passed
- root typecheck passed
- runtime production build passed

The PR remains Draft until the exact production reproduction passes against a real-provider 208 sandbox image built from this commit.

### DIFF
--- a/packages/runtime/src/__tests__/event-mapping.test.ts
+++ b/packages/runtime/src/__tests__/event-mapping.test.ts
@@ -6,6 +6,7 @@ import {
   type AgUiEvent,
 } from "@brainpilot/protocol";
 import { EventBus } from "../event-bus.js";
+import { RealAgentSession } from "../agent-factory.js";
 import { MasAgent } from "../mas-agent.js";
 import { MockAgentSession } from "../mock-agent.js";
 import { ev } from "../events.js";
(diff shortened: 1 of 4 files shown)


HaoxuanLiTHUAI:
Real-provider production reproduction now passes on 208 from exact commit `eeb525105f42d3a61f99120edf0b132e76e0a123`:

- candidate CPU sandbox built directly from this commit
- `deepseek-v4-flash`, reasoning `high`
- send obsolete 100-item QC task → Stop → immediately send `只回复：可以继续。`
- recovery completed and runtime returned idle in 2.5s
- no stale report text, no tool events, no Trace agent, no post-Stop file write

Final local gates are also green: root 99 files / 1112 tests, full workspace + docs build, typecheck, runtime 65 files / 671 tests. CI and CodeQL are green.

lucasli1015 (approved):
Reviewed the Stop-context rollback against #532 and the earlier #522/#527 queue fix. The checkpoint is captured only for non-streaming top-level prompts; follow-ups cannot replace it. Abort fences the active run, clears both queues, branches persisted Pi history, and rebuilds the live Agent message state before recovery. Unit, runtime, root, CI/CodeQL, and exact real-provider 208 evidence are green. Approved.

PR #555, opened 2026-09-08
fix(docker): make docs builds and GPU checks reproducible

HaoxuanLiTHUAI:
Building `docker/main/Dockerfile` after a host documentation build copies generated Fumadocs/Next caches into the image. Those caches contain the host's absolute `.source/source.config.mjs` path, causing the container build to fail resolving MDX files outside its filesystem root.

Exclude only the docs workspace's `.source`, `.next` and `out` directories from the Docker context. Documentation source remains included and is regenerated inside the builder. The paired GPU documentation checks also use `python3`, the executable verified in the published-base-derived GPU image. No runtime code, dependency, version or published tag changes.

Validation: reproduced on the Linux release checkout after a full host build; the same main builder now passes with all English/Chinese docs exports present. A real NVIDIA A10 tensor check passed with `python3`. The full Linux/amd64 main image now builds successfully from current head c69474a6 with the standard release proxy arguments; DataLad 1.6.0, DANDI 0.76.7 and wget checks pass. Actual dataset/container restart smoke follows before release acceptance.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,6 +4,10 @@ node_modules
 **/dist
 *.tsbuildinfo
 **/*.tsbuildinfo
+# Fumadocs/Next caches contain host-specific paths; regenerate them in the image.
+packages/docs/.source
+packages/docs/.next
+packages/docs/out
 .git
 .gitignore
 .worktrees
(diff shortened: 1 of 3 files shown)


lucasli1015 (approved):
Reviewed the narrow Docker build-context exclusions and both corrected GPU command examples. Generated .source/.next/out content is excluded while source remains included; Fumadocs now regenerates paths inside the image. Both the builder and full main image succeeded on Linux/amd64 after the host docs build, using the official Dockerfile and existing release proxy arguments. A real A10 CUDA tensor check succeeded with python3. Current CI/CodeQL passed. No runtime/dependency/version change is introduced.

Transparency: this is an AI-assisted review through the explicitly authorized alternate account; the same assistant prepared the change. It is not independent human review or publication/deployment approval.

Issue #135, opened 2026-06-20
Built-in skills tool name should match persona prompts


HaoxuanLiTHUAI:
## Summary

The built-in skills MCP tool should have a prompt-visible/callable name that matches the agent personas.

Right now the personas instruct agents to call:

```text
skills_tool_local
```

But the generic MCP bridge namespaces MCP tools as:

```text
mcp__<server>__<tool>
```

The built-in skills server is registered as `bp_skills_local`, so the actual exposed tool name may be:

```text
mcp__bp_skills_local__skills_tool_local
```

This mismatch likely makes agents less sensitive to skills usage: the prompt tells them to call the short name, while the tool list may expose the long namespaced one.

## Why this matters

Skills are now intended to be a core methodology layer for PI, writer, experimentalist, engineer, etc. If the tool name in the prompt does not match the actual callable tool name, agents may skip skills or attempt an invalid tool call.

## Expected behavior

The built-in BrainPilot skills tool should be easy for agents to call and should match the prompt text.

Preferred behavior:

- Built-in skills should be exposed as the bare tool name `skills_tool_local`.
- External MCP tools can keep namespacing (`mcp__server__tool`) to avoid collisions.
- If a collision occurs, the built-in tool should either reserve the bare name or use a clear documented alias.

## Possible implementation

Options:

1. Special-case the built-in skills bridge so it returns `skills_tool_local` instead of `mcp__bp_skills_local__skills_tool_local`.
2. Add alias support in `McpBridge` for trusted/built-in servers.
3. If namespacing must remain, update every persona prompt to reference the real namespaced tool name. This is less desirable because it leaks transport details into agent instructions.

Relevant files:

- `packages/runtime/src/session-manager.ts` starts the built-in skills MCP server as `bp_skills_local`.
- `packages/runtime/src/mcp-bridge.ts` wraps all MCP tools as `mcp__${server}__${tool}`.
- `packages/runtime/src/personas.ts` instructs agents to use `skills_tool_local`.

## Priority

Medium. This affects how reliably agents use the new skills system.

bondingelectron:
PR #141（`feat/native-skills-pipeline`）已彻底重构了技能加载架构，该 issue 描述的问题不再存在。

## 架构变更

原方案（PR #126）将技能作为 MCP stdio 工具暴露，导致 `mcp__<server>__<tool>` 命名空间包装与 persona 提示词中的短名不一致。PR #141 完全替换了这一机制：

[...]

PR #562, opened 2026-09-15
fix(web): clarify Settings and desktop UX states

HaoxuanLiTHUAI:
## Summary
This ordinary functional-fix PR now includes the originally reviewed Settings fix plus the selected desktop UX follow-up. It is **not** a version/release PR.

- Load Settings provider/MCP/plugin resources independently; retain successful data when another endpoint fails. Respect shared-provider readonly permissions and explicit hosted capabilities.
- Distinguish pending, unavailable, confirmed-empty and cached states in the composer, session list, search and history; provide scoped retries without losing usable data.
- Keep saved/manual/pinned provider and model choices stable across late responses and session changes.
- Deduplicate replayed terminal errors by stable event aliases, including rich-diagnostic/history hydration merges; preserve retry/edit recovery controls.
- Keep Resources navigation capability-aware, display primary-view labels and readable conversation titles, and remove redundant visible IDs.
- Synchronize file selection/close/reference/deletion with the URL, reject stale async previews and preserve the existing single unsaved-changes confirmation.
- Show concise Provider/Trace/Overview/File summaries with complete optional details; keep failures, artifacts and pending decisions prominent. Localize metadata and correct file-retention wording.
- Retain retry controls through pending state and avoid search-focus resets on unrelated renders.

## Source and scope
- Settings commit: `7cf756f5ee3f36ec12e6774ef826f088b41b5529`.
- Desktop UX commit: `11829d9ab6e6c47f1684a5e1d8639ef1b5ada0e0` (46 additional Web source/test files).
- No mobile redesign, version/dependency/lockfile change, backend schema migration, package/image publication or deployment.
- Cloud companion #135 is already merged, but production still requires a new Web artifact plus the separate release acceptance/rollout gates.

## Verification
- Reviewed desktop source manifest matched byte-for-byte before commit.
- Web: 904 tests across 99 files passed; TypeScript and production build passed on those exact source bytes.
- Root TypeScript and read-only version alignment check passed.
- Real isolated local HTTP + mock-runtime checks covered provider state, model persistence, error replay/reload, stopped-backend recovery, compact provider cards, and file selection/URL/localized metadata.
- Exact-head GitHub CI run **35194652414 passed** for `d77ff8aa8eda95c886cc97bac217423c3848ab12`: install, root typecheck, non-Web tests, Web tests/build, docs checks/build and all-package build. Earlier Settings-only CI is not reused for this head.

## Explicit pending acceptance / local diagnostics
- The previously blocked desktop follow-up scenarios have now been exercised in an explicitly authorized, isolated headless Chrome instance against a real local backend with BP_MOCK=1 runtime. The new layout/focus findings were fixed and re-tested (details below). This is local desktop acceptance, not production or exact-main release acceptance.
- The local default-parallel non-Web run reported 41 failures (mostly 5-second timeouts); a one-worker run with unchanged timeouts reduced that to two timeouts in history/restore tests. Both full-run logs are retained. Isolating the two remaining files passed all 9 tests with the original 5-second timeout; no source/test-timeout changes were made. The failed local full runs are not described as green.
- No production inference or exact-main release acceptance is claimed. Existing full-image dependency/security inventory remains a separate release-readiness item; unrelated dependency PRs were not silently included.
- Required approving review and strict branch protection must be honored. No account-switch self-review or bypass.

## History loading ownership follow-up
- Commit `70a5791cccff993b337925f987324d3e084ecf92` fixes the new cached-session/draft loading regression identified in author-side verification.
- Pending reads are keyed by session and request sequence; completion releases only the entry it owns. Active-scope loading is derived, while existing stale-data/SSE guards remain unchanged.
- Added 5 regression cases (14 focused history-state tests total): cached B, draft, return to pending A, overlapping same-session refreshes, and abandoned-request failure.
- The original unchanged two-case diagnostic now passes. Full Web suite: 893/893; TypeScript/build/diff-check passed. New exact-head CI 35189264671 passed all required steps.
- This is fix verification, not an independent approving review. The subsequent headless desktop follow-up is documented below; no merge/publication/deployment has occurred.

## Headless desktop acceptance follow-up
- Commit `d77ff8aa8eda95c886cc97bac217423c3848ab12` fixes three remaining UX findings: title wrapping lost to CSS specificity; expanded Trace filters squeezed into an obsolete grid column; successful Settings/Search retries dropped focus onto body.
- Validated in isolated headless Chrome at 1280×720, with disposable copied test data and no real model inference or user-browser control. Native unsaved-cancel, file URL/reference/focus, provider details, history delayed-response races and reload dedup, Overview/Trace disclosures, and Settings/Search retry pending/failure/success were exercised.
- Real measurements: long titles now occupy 2 lines; expanded Trace controls height 254→124 px, graph top 488→358 px. Both successful-retry focus regressions reproduced before the patch and pass afterward.
- Full Web suite **904/904**, TypeScript/build/diff-check passed. Exact-head CI **35194652414** passed all required steps.
- The isolated browser and QA servers were stopped; the two file fixtures retained their original byte hashes, and raw history still contains one RUN_ERROR.
- No independent approval, merge, immutable artifact publication or production deployment is implied. Required review and the durable release lifecycle remain in force.


HaoxuanLiTHUAI:
## Author-side verification: changes needed before approval

This is a transparent self-review/verification note from the current author account, **not** an independent approving review. No account switch or permission change was used.

### [P2] A history refresh can leave another cached session permanently marked as loading

At `packages/web/src/contexts/SessionContext.tsx:1135-1136`, the new completion guard refuses to clear the global `isRefreshingMessages` flag after the user switches away from the refreshed session. That protects a newly loading session, but the selection effect at lines 593-598 immediately returns for an already-hydrated session or a draft. In those cases no new request owns or clears the flag.

Reproduction:
1. Open A and B so both histories are hydrated.
2. Return to A and start a manual history refresh; defer its response.
3. Switch to cached B (or start a new draft).
4. Resolve A's refresh.
5. `isRefreshingMessages` remains `true` even though no history request remains.

On cached B this leaves the “loading history” notice visible; if a history-error retry is rendered, the stale flag also leaves that retry busy/guarded. The toolbar refresh action is still available, so this is not a claim that all interaction is blocked.

**Controlled verification:** the same two regression cases pass on base `5a3a76bf7618e807345953e3be79597c3cf9e682` and both fail on head `11829d9ab6e6c47f1684a5e1d8639ef1b5ada0e0` (expected `false`, received `true`). They exercise the actual `SessionProvider` with deferred API promises, not a source-string assertion. The temporary tests did not modify this PR's source.

Suggested correction: track pending history reads by session/request identity and derive the active session's loading state, or explicitly handle cached-session/draft transitions while preserving the existing safeguard for a newly loading B. Do not simply remove the new guard, which would restore the previously tested A→uncached-B race.

The existing green CI does not cover these two transitions. Final browser acceptance also remains incomplete. My current recommendation is **changes needed**, not approval or deployment.

HaoxuanLiTHUAI:
### Loading-state review finding fixed in `70a5791cccff993b337925f987324d3e084ecf92`

The pending flag is now owned by `(sessionId, request sequence)` instead of one global boolean. Leaving A for cached B or a draft no longer carries A's busy state across; an older completion cannot clear a newer request. Existing message/error/SSE stale-response guards are preserved.

Verification:
- Original unchanged two-case reproduction: **2/2 passed** (previously both failed on 11829d9a).
- Focused history-state tests: **14 passed**, including 5 new regression cases.
- Full Web suite: **893 passed**; TypeScript, build and diff-check passed.
- Exact-head GitHub CI **35189264671 passed**, including non-Web tests, Web tests/build, docs checks and all-package build.

The change is limited to two files, with no version/dependency changes. This author-side verification does **not** count as independent approval. Final real desktop browser acceptance is still pending; no merge or deployment was performed.

HaoxuanLiTHUAI:
### Headless desktop follow-up complete; fixes in `d77ff8aa8eda95c886cc97bac217423c3848ab12`

Using a separate, authorized headless Chrome instance and disposable local data, I reproduced and fixed:
1. The two-line title rule lost to the generic span nowrap rule. Titles now render two lines.
2. Trace's new filter disclosure still occupied an old narrow grid column. Expanded controls now span the row; measured height fell from 254 to 124 px and the graph starts at y=358 rather than y=488 (1280×720).
3. Successful Settings/Search retries removed the focused Retry button before focus could be restored. A shared post-commit hook now restores the section heading/search input, while preserving focus during pending/failure and not stealing it from another control.

The previously blocked file close/reference/unsaved-cancel flows, native disclosures, history A→cached-B/draft races, reload error deduplication, and retry focus paths were exercised; affected cases were rerun after rebuilding. Full Web suite: **904 passed**, plus TypeScript/build/diff-check. Exact-head CI **35194652414 passed**.

No production or real-provider inference tests were substituted by these local mock-runtime checks. This remains author-side verification, **not an independent approving review**. No merge or deployment was performed.

lucasli1015 (commented):
## Astra/Opus read-only review — 2026-09-17 — d77ff8aa

**COMMENT-only automated review, not an independent approval.** This operator-requested review was generated by the configured Opus workflow (`custom-model-a9`) against immutable base/head snapshots, then its findings were checked by Astra. It is submitted using an operator-controlled account. It does not approve, merge, publish, or replace required independent review.

Reviewed head: `d77ff8aa8eda95c886cc97bac217423c3848ab12`  
Base: `5a3a76bf7618e807345953e3be79597c3cf9e682`

### Result after verification
Opus found no confirmed new P1/P2 regression in its source-review scope and raised two P3 items. Astra reproduced the underlying data/state behaviors and checked whether their preconditions exist in the shipped application. The resulting assessment is **one conditional legacy-compatibility finding and one defensive-hardening suggestion**, not two demonstrated current-production failures.

### [P3] Legacy shared-provider fallback is lost after normalization
`packages/web/src/components/settings/providerSettingsView.ts:18-31` and `packages/web/src/contracts/backend.ts:860`.

For an older/skewed server that omits both `isShared` and `is_shared`, `normalizeProviderProfile` turns the missing field into `false`. The view helper consequently never reaches its documented `shared_` prefix fallback. A normalized legacy shared profile is grouped as private and offered Edit/Remove. Base Settings grouped by prefix, so this narrowly scoped compatibility behavior changed.

Astra's direct normalizer → permissions test confirms the gap: the missing-field `shared_legacy` case allows editing instead of remaining readonly. Controls for explicit `isShared: true` with an opaque id and explicit `isShared: false` with a `shared_` id both pass. The raw helper's missing-field fallback also passes, demonstrating why its existing isolated tests do not exercise the real API-normalization path.

**Qualification:** the current protocol requires `isShared`; we have not demonstrated that current Cloud responses omit it. This is a conditional compatibility issue, not a claim that current platform profiles are all misclassified or that a particular legacy server's write endpoint was tested.

**Suggested disposition:** if missing-field legacy servers remain supported, apply the fallback at the raw-normalization boundary while preserving explicit `false`, and add an end-to-end normalization/view test. Otherwise remove the unreachable compatibility promise/tests and document the minimum server contract.

### Defensive-hardening suggestion: clear list loading on a not-auth-ready reset
`packages/web/src/contexts/SessionContext.tsx:558-609`, especially the early reset at 559-567 and sequence-guarded completion at 607.

If a pending session-list request is superseded by a hypothetical `isAuthReady: true → false` change, the reset increments the request sequence and returns without clearing `isLoading`; the old completion can no longer clear it. A deferred-request test with an injected auth reset reproduces the stale flag.

**Important correction to the initial Opus classification:** the shipped `AuthProvider` initializes readiness to false and only sets it to true (`contexts/AuthContext.tsx:22,32,40`); it exposes no readiness setter or sign-out transition back to false. Thus this is not a demonstrated current logout/token-invalidation bug. Current UI also uses the explicit list status. Treat it as low-priority hardening if an in-place auth reset is later introduced, not as a present-day user-visible blocker.

### Scope, evidence, and limits
- Opus read all three production diff groups (state/API, providers/Settings, navigation/presentation), with targeted HEAD/base source and test inspection. It had only Read/Glob/Grep tools, no file writes, shell, browser, network, GitHub credentials or mutation tools.
- Astra's five targeted verification cases: three control cases passed; the legacy-compatibility expectation and hypothetical auth-reset expectation failed as described above. Temporary test code was removed from the working tree; no product code changed.
- Opus's replay-merge and draft-send-gate questions remain unconfirmed hypotheses, not additional defects. The no-event-id terminal-card behavior it mentioned also exists in base.
- The existing 904-test CI/headless evidence was context, not tests run by Opus. This source pass is not an exhaustive security, compatibility, or production audit.
- Required review, exact-main verification, release-surface compatibility/security decisions and publication/deployment approvals remain separate.

HaoxuanLiTHUAI:
### Maintainer triage — 2026-09-18

Per the user's decision, the conditional legacy-provider compatibility item and hypothetical auth-reset hardening suggestion from the automated review are **deferred / non-blocking for this PR**. No additional code change for those P3 items is requested here.

The current fix candidate remains `d77ff8aa8eda95c886cc97bac217423c3848ab12`. A separate broader P1/P2 UX investigation is being tracked independently rather than silently expanding this PR's scope.

This comment records the maintainer's technical disposition; it is not an approving review and does not change protected-branch requirements.

